PR #2, opened 2014-12-25
Caches and nodes filtering + e2e tests

tsykora:
Christmas PullRequest!! :)

Hello guys,
I've implemented filtering for caches and nodes at particular cluster's page. Considering deployment, where there live more than 100 caches, that could be useful. 

Also I've added e2e protractor tests for all implemented behaviour. Working on my local machine, hope that also works for you (probably u need Chrome, so far) and also running Infinispan server (of course). 

Let me know if you spot something. 

Note: I had to change Cluster.caches (cluster-model-service.js) variable from an Object ({}) to an Array ([]) because filtering in AngularJS works only for arrays right now. There is an open issue for ng to make it work also for Objects. This shouldn't cause any problems for us now. It's just an array.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,7 @@ dist
 .settings
 .project
 .classpath
+.idea
+infinispan-management-console.iml
 src/main/bower_components
 target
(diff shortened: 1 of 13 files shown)


tsykora:
Surpassed by: https://github.com/infinispan/infinispan-management-console/pull/3
Closing.

PR #3, opened 2015-01-01
Cache filter and REST endpoint connection

tsykora:
Hello!!
This is New Year's Day pull request. I want to say my PF 2015 this way to everyone involved. 

This PR surpass former Christmas PR: https://github.com/infinispan/infinispan-management-console/pull/2 and instead of cache filtering is adding an experimental UI for communicating with Infinispan REST server. 

Putting and getting entries to/from caches.

I've spent a couple of hours figuring out how to deal with/bypass same-origin-policy stuff. The simplest sollution (IMHO) is to use corsproxy locally. This experimental functionality can be used for testing, etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,7 @@ dist
 .settings
 .project
 .classpath
+.idea
+infinispan-management-console.iml
 src/main/bower_components
 target
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ dist
 infinispan-management-console.iml
 src/main/bower_components
 target
+README.md~
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,4 +2,5 @@ Martin Šošić
 Matija Šošić
 Tristan Tarrant
 Vladimir Blagojevic
+Tomáš Sýkora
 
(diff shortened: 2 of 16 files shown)


tristantarrant:
I'd rather implement the cache content manipulation in a different manner (i.e. not endpoint dependent), possibly using the interpreter ops (which is what the CLI uses).

tsykora:
Ok, thanks Tristan, I will look on that soon.

It was just something what came to my mind and I wanted to play with it a bit.

tsykora:
Trying to get proof of concept working, but getting:

{
    "outcome" : "failed",
    "result" : null,
    "failure-description" : "JBAS014884: No operation named 'put' exists at address [\n    (\"subsystem\" => \"infinispan\"),\n    (\"cache-container\" => \"clustered\"),\n    (\"distributed-cache\" => \"default\")\n]",
    "rolled-back" : true
}

this is strange, because put is enlisted in help --commands under particular cache when logged at server using jboss-cli.sh. When I try the same with operation read-attribute -- it works ok (and it's already there implemented). 
What I don't know now is why I can't call PUT operation.

Any ideas? Thank you! :smile:

tsykora:
I've got my answer... PUT operation is not present when started in domain mode. 

We have to implement management for server started using standalone.sh. Then I believe it will work. Will try to implement a POC.

tsykora:
Closing, we need to find easier and better sollution.

PR #4, opened 2015-02-21
Cache filtering login

tsykora:
:) I've implemented filtering for caches and nodes at particular cluster's page. Considering deployment, where there live more than 100 caches, that could be useful.

I've also added e2e protractor tests for all newly implemented behaviour.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,7 @@ dist
 .settings
 .project
 .classpath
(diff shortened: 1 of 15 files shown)


tsykora:
surpassed by: https://github.com/infinispan/infinispan-management-console/pull/5 closing...

PR #8, opened 2015-03-31
ISPN-5234 Implement single cluster view page (P1)

vblagoje:
Cumulative PR for all features listed in https://issues.jboss.org/browse/ISPN-5234

### DIFF
--- a/src/main/webapp/cluster-view/cluster-view.controller.js
+++ b/src/main/webapp/cluster-view/cluster-view.controller.js
@@ -33,6 +33,27 @@ angular.module('managementConsole')
         return $scope.shared.currentCollection === 'caches';
       };
 
+      $scope.zoomedIn = false;
+
+      var zoomChanged = function () {
+        var slid = $("#zoom");
+        var zoomValue = slid.val();
+        $scope.zoomedIn = parseInt(zoomValue) === 1 ? true : false;
+        $scope.$apply();
+      };
+
+      var zoomSliderElement = $("#zoom");
+      zoomSliderElement.noUiSlider({
+        start: [0],
+        step:1,
+        range: {
+          'min': [0],
+          'max': [1]
+        }
+      });
+      zoomSliderElement.Link('lower').to($("#zoom-value"));
+      zoomSliderElement.on('change', zoomChanged);
+
       // define sliders
       var sliderNames = ['average-read-time', 'average-write-time', 'average-remove-time', 'average-replication-time'];
 
--- a/src/main/webapp/cluster-view/cluster-view.controller.js
+++ b/src/main/webapp/cluster-view/cluster-view.controller.js
@@ -5,8 +5,9 @@ angular.module('managementConsole')
     '$scope',
     '$stateParams',
     '$state',
+    '$q',
     'modelController',
-    function ($scope, $stateParams, $state, modelController) {
+    function ($scope, $stateParams, $state, $q, modelController) {
       if (!modelController.isAuthenticated()) {
         $state.go('/logout');
       }
@@ -24,31 +25,66 @@ angular.module('managementConsole')
         }
       });
 
+      $scope.isNodesView = function (){
+        return $scope.shared.currentCollection === 'nodes';
+      };
+
+      $scope.isCachesView = function (){
+        return $scope.shared.currentCollection === 'caches';
+      };
+
       // define sliders
       var sliderNames = ['average-read-time', 'average-write-time', 'average-remove-time', 'average-replication-time'];
 
       $scope.clearFilters = function () {
+
+        //reset all sliders to initial values
         for (var i = 0; i < sliderNames.length; i++) {
           var slider = $("#" + sliderNames[i]);
           slider.val([0, 60000]);
         }
+
+        //show all caches
         var cachesMap = $scope.currentCluster.getCaches();
         for (var k in cachesMap) {
           var cache = cachesMap[k];
           cache.show = true;
         }
-      }
 
-      var filterOutCaches = function () {
+        //show all nodes
+        var server = modelController.getServer();
+        var nodes = server.getNodes();
+        for (var i = 0; i < nodes.length; i++) {
+          nodes[i].show = true;
+        }
+      };
+
+      var filterView = function () {
+        var filterChangedName = $(this).context.id;
+        if ($scope.isCachesView()) {
+          filterOutCaches(filterChangedName);
+        } else if ($scope.isNodesView()) {
+          filterOutNodes(filterChangedName);
+        }
+      };
+
+      var filterOutCaches = function (changedFilter) {
         var cachesMap = $scope.currentCluster.getCaches();
         for (var k in cachesMap) {
           // do checks
-          var filterChangedName = $(this).context.id; //metric i.e. statistic we are checking
-          var slid = $("#" + filterChangedName);
+          var slid = $("#" + changedFilter);
           var lowerValue = slid.val()[0];
           var upperValue = slid.val()[1];
-          filterCacheVisibility(filterChangedName, cachesMap[k], lowerValue, upperValue);
+          filterCacheVisibility(changedFilter, cachesMap[k], lowerValue, upperValue);
         }
+      };
+
+      var filterOutNodes = function (changedFilter) {
+        // do checks
+        var slid = $("#" + changedFilter);
+        var lowerValue = slid.val()[0];
+        var upperValue = slid.val()[1];
+        filterNodeVisibility(changedFilter, lowerValue, upperValue);
       };
 
 
@@ -80,7 +116,7 @@ angular.module('managementConsole')
         // e.g.: slider-avg-write-time-value-lower
         sliderElement.Link('lower').to($("#" + sliderNames[i] + "-value-lower"));
         sliderElement.Link('upper').to($("#" + sliderNames[i] + "-value-upper"));
-        sliderElement.on('change', filterOutCaches);
+        sliderElement.on('change', filterView);
       }
 
       var filterCacheVisibility = function (statisticName, cache, lowerValue, upperValue) {
@@ -88,6 +124,26 @@ angular.module('managementConsole')
         p.then(function (response) {
           var statValue = response[0][statisticName];
           cache.show = (lowerValue <= statValue && statValue <= upperValue);
+        });
+      };
+
+      var filterNodeVisibility = function (statisticName, lowerValue, upperValue) {
+        var server = modelController.getServer();
+        var nodes = server.getNodes();
+        var promises = [];
+        for (var i = 0; i < nodes.length; i++) {
+          promises.push(server.fetchAggregateNodeStats($scope.currentCluster, nodes[i]));
+        }
+        var p = $q.all(promises);
+        p.then(function (responses) {
+          var resp = undefined;
+          for (var i = 0; i < responses.length; i++) {
+            resp = responses[i];
+            if (resp) {
+              var statValue = resp[statisticName];
+              nodes[i].show = (lowerValue <= statValue && statValue <= upperValue);
+            }
+          }
         });
       };
     }]);
--- a/src/main/webapp/cluster-view/cluster-view.controller.js
+++ b/src/main/webapp/cluster-view/cluster-view.controller.js
@@ -1,26 +1,93 @@
 'use strict';
 
 angular.module('managementConsole')
-    .controller('ClusterViewCtrl', [
+  .controller('ClusterViewCtrl', [
     '$scope',
     '$stateParams',
     '$state',
     'modelController',
     function ($scope, $stateParams, $state, modelController) {
-            if (!modelController.isAuthenticated()) {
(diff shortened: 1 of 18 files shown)


vblagoje:
@tristantarrant would you please add Martinsos (sosic.martin@gmail.com) to the project so he can review this PR

vblagoje:
@Martinsos this is the PR that needs review! Thank you!

src/main/webapp/cluster-view/cluster-view.controller.js
```diff
@@ -1,26 +1,170 @@
...
     '$scope',
     '$stateParams',
     '$state',
+    '$q',
     'modelController',
-    function ($scope, $stateParams, $state, modelController) {
-            if (!modelController.isAuthenticated()) {
-                $state.go('/logout');
+    function ($scope, $stateParams, $state, $q, modelController) {
+      if (!modelController.isAuthenticated()) {
```
Martinsos: I am reviewing my old code here :), but this is not the best practice: here we first load the controller, view will probably partially render, then we check if user is authenticated and if not, we send him to logout. As a result, we sometimes get flashing of view that we were not supposed to see.

Better practice is not to even load controller if user is not authenticated! That is usually done by checking conditions in main module, by listening to '$stateChangeStart' event -> there we can check if state is public or private, and if it is public, we can check if user is authenticated: if not, we can return user to login, and if yes, we can forward him to wanted destination. I suggest creating a separate issue for this one. Also, this way there is no need to check authentication in each controller, so there is no code duplication.

src/main/webapp/cluster-view/cluster-view.controller.js
```diff
@@ -1,26 +1,170 @@
...
+      };
+
+      $scope.isCachesView = function (){
+        return $scope.shared.currentCollection === 'caches';
+      };
+
+      $scope.zoomedIn = false;
+
+      var zoomChanged = function () {
+        var slid = $("#zoom");
```
Martinsos: Generally, there should be no DOM manipulation in angular controller: controller should contain only bussiness logic, basically only connecting other components. All DOM manipulation should be done in directives, and shared logic should be done in services. Here, we have DOM manipulation, so we should consider moving it to directive. Same applies to sliders below. I was not able to run the app yet, but will comment more on this when I get it running.

Martinsos: Ok I finally have the thing running! I took a look at interface, and if I am correct, the whole filtering goes like this: we have certain values (average read time, write time and so on) and we have sliders which modify this values. When values modify, we do adjust `show` attribute of nodes/caches.

Slider is a highly reusable component with view (html) and logic (js), and basically all it does is output a pair of integers. So instead of having all this slider logic in this controller, we should make it a directive, and we would use it something like this in `cluster-view.html`:
```
<slider ng-model="avgWriteTime" max="maxAvgWriteTime" min="minAvgWriteTime"></slider>
<slider ng-model="avgReadTime" max="maxAvgReadTime" min="minAvgReadTime"></slider>
...
```
Then, to trigger filtering based on changes of filter values, we would use `$watch` in controller to adjust the `show` attributes of nodes/caches. Awesome thing here is that we do not really care who is changing filter values: it can be sliders or anything else! So basically all goes around filter values: we have sliders changing them, and controller watching them and reacting to change by doing filtering.

After this change there should be much less code in the controller.

About slider directive: you can either find some existing angular directive (like https://github.com/seiyria/angular-bootstrap-slider that was used so far -> why are we not using it anymore?), or you can wrap ui-slider in your own directive, from what I see it should be really easy to do, basically you create a new directive, move the UiSlider code there to the right places, expose attributes and that is it -> good opportunity to learn more about directives.

src/main/webapp/cluster-view/cluster-view.controller.js
```diff
@@ -1,26 +1,170 @@
...
+      var filterNodeVisibility = function (statisticName, lowerValue, upperValue) {
+        var server = modelController.getServer();
+        var nodes = server.getNodes();
+        var promises = [];
+        for (var i = 0; i < nodes.length; i++) {
+          promises.push(server.fetchAggregateNodeStats($scope.currentCluster, nodes[i]));
+        }
+        var p = $q.all(promises);
+        p.then(function (responses) {
+          var resp = undefined;
```
Martinsos: There is no need for undefined here, it is undefined by default.

src/main/webapp/cluster-view/cluster-view.controller.js
```diff
@@ -1,26 +1,170 @@
...
+          for (var i = 0; i < responses.length; i++) {
```
Martinsos: For-loop in javascript does not have its own scope as in other languages, so in some cases unexpected things can happen if you are not aware of that: therefore I prefer using `angular.forEach` when possible, and I would use it here instead of for loop! Code looks nicer, and is is harder to make a mistake.

src/main/webapp/cluster-view/cluster-view.html
```diff
@@ -41,8 +41,13 @@
 
     <!-- TOP ROW #2 START -->
     <div class="row">
-        <div class="col-md-2 text-center">
-            <h4 style="color:   #009A31;">Available</h4>
+        <div class="col-md-2 text-left">
+            <div ng-show="currentCluster.availability === 'AVAILABLE'">
```
Martinsos: Instead of duplicating html here, we can do it a little bit better: remove the ng-show, and directly modify properties according to availability. So instead of `class="fa fa-1x unavailable"` you would use `class="fa fa-1x" ng-class="{available: currentCluster.availability === 'AVAILABLE', unavailable: currentCluster.availability === 'UNAVAILABLE'}"`, and {{currentCluster.availability}} as paragraph content, instead of hardcoded value.

src/main/webapp/cluster-view/cluster-view.html
```diff
@@ -78,27 +86,35 @@ <h4 style="color:   #009A31;">Available</h4>
...
-                    <span class="slider-text"> Average read time </span> 
-                    <br>
-                    <div class="slider-one"></div>
-                    <span class="slider-min"> 100 </span>  <span class="slider-max"> 200 </span> 
                     <br>
                     <br>
+                    <span>Average remove time</span>
+
+                    <div id="average-remove-time"></div>
+                    <span class="slider-min" id="average-remove-time-value-lower"></span>
```
Martinsos: Here you would use slider directive!

src/main/webapp/cluster-view/cluster-view.html
```diff
@@ -126,28 +142,46 @@ <h4 style="color:   #009A31;">Available</h4>
...
-                        <a class="text-center" ui-sref="nodeDetails({clusterName: currentCluster.name, nodeName: node.name})">
-              {{node.name}} <br><br>
-            </a>
+                    <div class="box col-md-2 nopadding"
+                         ng-repeat="node in currentCluster.getNodes() | customPropertyFilter:'name':searchQuery"
+                         ng-show="node.show">
+                        <a class="text-center"
+                           ui-sref="nodeDetails({clusterName: currentCluster.name, nodeName: node.name})">
+                            {{node.name}} <br>
+                            <h6 ng-show="zoomedIn">5 caches</h6>
```
Martinsos: Not so important, but I would maybe wrap these two header in `<div>` or `<span>` and use ng-show only on it, it will make more sense to do it if we there will be more values later.

src/main/webapp/cluster-view/jquery.nouislider.ispn.css
```diff
@@ -0,0 +1,166 @@
+
```
Martinsos: This is vendor css, correct? In that case we should place it somewhere else! If you create directive using UI-slider you could put this css in same directory with directive.

src/main/webapp/components/api/cache-model.service.js
```diff
@@ -11,6 +11,7 @@ angular.module('managementConsole.api')
                 this.modelController = cluster.getModelController();
                 this.lastRefresh = null;
                 this.data = null;
+                this.show = true;
```
Martinsos: This is attribute specific for view, not for data, so it does not feel completely right for CacheModel to know about it. Few ideas: rename it to `hide`, that way by default model is shown, and there is no need to set it here: it is better because now Model does not have any knowledge about view logic, but bad side that still remains is that view pollutes model with its own data.

Best approach would probably be the following: instead of using attribute that describes if node/cache is hidden or not, you could calculate it on the fly: in controller, instead of checking for `ng-show="node.show"`, you could do `ng-show=isNodeShown(node)`, where `isNodeShown` would be your function that checks based on filters values and node values if node should be shown or not. In that case you do not even need to use '$watch' to react to filter values changes. I would say this is certainly the best way regarding the beauty of logic and code, only argument against would be performance: if there are many nodes and caches (for example hundred and more) then it may affect performance, also performance may be affected if it is complicated to fetch cache/node stats (although I believe rest calls are cached right? Or do we expect these stats to change so we can not cache it?).

src/main/webapp/components/filters/property-filter.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+angular.module('managementConsole.filters', [])
+    .filter('customPropertyFilter', function (){
```
Martinsos: File is named property-filter, so I would probably name the filter the same way: word "custom" is not really descriptive anyway, it does not bring much value I think.

Martinsos: Also, it would be nice to write some documentation here! This filter is important because it can filter deep properties, so that should be stated, and argument should be explained.

src/main/webapp/index.html
```diff
@@ -11,7 +11,11 @@
     <link rel="stylesheet" type="text/css" href="vendor.css">
     <!-- bower:css -->
     <link rel="stylesheet" href="bower_components/patternfly/dist/css/patternfly.css" />
-    <link rel="stylesheet" href="bower_components/seiyria-bootstrap-slider/dist/css/bootstrap-slider.css" />
+    <link rel="stylesheet" href="bower_components/nouislider/distribute/jquery.nouislider.min.css" />
+    <!-- slight modifications to fit our style and sizes -->
+
+    <!-- TODO: delete these links probably, we use nouislider -->
+    <!--<link rel="stylesheet" href="bower_components/seiyria-bootstrap-slider/dist/css/bootstrap-slider.css" />-->
```
Martinsos: Yes we can delete this!

src/main/webapp/index.html
```diff
@@ -50,8 +54,12 @@
     <script src="bower_components/angular-ui-router/release/angular-ui-router.js"></script>
     <script src="bower_components/angular-ui-bootstrap-bower/ui-bootstrap-tpls.js"></script>
     <script src="bower_components/patternfly/dist/js/patternfly.js"></script>
-    <script src="bower_components/seiyria-bootstrap-slider/dist/bootstrap-slider.min.js"></script>
-    <script src="bower_components/angular-bootstrap-slider/slider.js"></script>
+    <script src="bower_components/nouislider/distribute/jquery.nouislider.all.js"></script>
+
+    <!-- TODO: delete these links probably, we use nouislider -->
+    <!--<script src="bower_components/seiyria-bootstrap-slider/dist/bootstrap-slider.min.js"></script>-->
```
Martinsos: we can delete this also if we will stick to nouislider.

src/main/webapp/index.html
```diff
@@ -71,8 +79,13 @@
     <script src="components/api/profile-model.service.js"></script>
     <script src="components/api/server-model.service.js"></script>
     <script src="components/api/servergroup-model.service.js"></script>
+    <script src="components/filters/property-filter.js"></script>
     <script src="components/navbar/navbar.controller.js"></script>
- 
+
+    <link rel="stylesheet" href="cluster-view/jquery.nouislider.ispn.css" />
```
Martinsos: We usually put stylesheets to `<head>`!

src/main/webapp/node-details/node-details.html
```diff
@@ -23,7 +23,9 @@
     <div class="node-details">
         <div class="node-name">{{currentNode.name}}</div>
         <div class="node-status">
-            <span class="node-status-label">Status:</span>  <span class="node-status-value available">{{currentNode.state}}</span>
+            <span class="node-status-label">Status:</span>
+            <span ng-show="currentNode.isRunning()" class="node-status-value available">{{currentNode.state}}</span>
```
Martinsos: Same thing as before, we can reuse html here and not repeat it.

Martinsos:
@vblagoje Nice job for the first angular app :)! I wrote some suggestions and ideas, when we are finished with them I can review it once more, because structure will probably significantly change.

tsykora:
@Martinsos thank you very much for all your suggestions!

tsykora:
@vblagoje please, don't force push over Martin's comments. I'd like to keep them here in closed PR please!

vblagoje:
Sure thing @tsykora . @Martinsos Your input was very valuable. Much appreciated.

PR #9, opened 2015-04-10
ISPN-5234 Implement single cluster view page (P1)

vblagoje:
Master only. See https://issues.jboss.org/browse/ISPN-5234 for more details. @Martinsos please review once again


vblagoje:
@Martinsos I believe I have addressed all your comments from https://github.com/infinispan/infinispan-management-console/pull/8 However, I have not implemented cache/node show completely in view. I will look at that once I understand Angular a bit more. Thanks once again.

vblagoje:
@tristantarrant @tsykora Please have a look and try out the app as well. After Martin's suggestions it was not too hard to polish this one so it looks awesome! I used Angular design patterns as much as possible. It was so much fun!

src/main/webapp/cluster-view/cluster-view.controller.js
```diff
@@ -1,26 +1,150 @@
...
     '$scope',
     '$stateParams',
     '$state',
+    '$q',
     'modelController',
-    function ($scope, $stateParams, $state, modelController) {
-            if (!modelController.isAuthenticated()) {
-                $state.go('/logout');
+    function ($scope, $stateParams, $state, $q, modelController) {
+      if (!modelController.isAuthenticated()) {
```
Martinsos: @vblagoje: we should not check if user is authenticated here because it is too late then, controller will already start rendering and there will be flashing and similar which is not nice. We should listen to state change instead ($stateChangeStart), and on each state change trigger logic that will check if user has permission to go that state or not.

Martinsos:
I started reviewing and will finish it tonight (I will write a comment when I am done)!

src/main/webapp/cluster-view/cluster-view.controller.js
```diff
@@ -1,26 +1,150 @@
...
+
+      $scope.$watch('currentCluster', function (currentCluster) {
+        if (currentCluster && currentCluster.name !== $stateParams.clusterName) {
+          $state.go('clusterView', {
+            'clusterName': currentCluster.name
+          });
+        }
+      });
+
+      $scope.currentClusterState = function () {
```
Martinsos: This are really details, but maybe instead of implementing all this `currentCluster` functions, it may be nicer to implement this functions for any cluster (either as cluster methods or as function that takes cluster object) and then use them on current cluster, what do you think?

src/main/webapp/cluster-view/cluster-view.controller.js
```diff
@@ -1,26 +1,150 @@
...
+        },
+        {
+          name: "Average replication time",
+          id: 'average-replication-time',
+          from: $scope.initialMin,
+          to: $scope.initialMax
+        }
+      ];
+
+      angular.forEach($scope.sliders, function (slider) {
```
Martinsos: While this is ok, even more cool approach would be to implement `onChange` attribute for slider that would execute given expression when slider value changes, very similar to what `ng-change` does, so you would not have to do watch here.

src/main/webapp/cluster-view/cluster-view.controller.js
```diff
@@ -1,26 +1,150 @@
...
+      var filterOutNodes = function (sliderArray) {
+        var server = modelController.getServer();
+        var nodes = server.getNodes();
+        var promises = [];
+        angular.forEach(nodes, function (node) {
+          promises.push(server.fetchAggregateNodeStats($scope.currentCluster, node));
+        });
+        var p = $q.all(promises);
+        p.then(function (responses) {
+          for (var i = 0; i < responses.length; i++) {
```
Martinsos: Instead of for loop, use `angular.forEach` -> it is less prone to mistakes, and it is also more clear what is happening: you are iterating through all elements.

src/main/webapp/cluster-view/cluster-view.html
```diff
@@ -41,14 +41,17 @@
 
     <!-- TOP ROW #2 START -->
     <div class="row">
-        <div class="col-md-2 text-center">
-            <h4 style="color:   #009A31;">Available</h4>
+        <div class="col-md-2 text-left">
+            <p class="fa fa-1x"
+               ng-class="currentClusterAvailability()?'available':'unavailable'">
```
Martinsos: Lets put spacing around ? and : ! No need to save space here.

src/main/webapp/cluster-view/cluster-view.html
```diff
@@ -126,28 +122,52 @@ <h4 style="color:   #009A31;">Available</h4>
...
-                <div ng-show="shared.currentCollection == 'nodes'">
-                    <div ng-show="!currentCluster.getNodes() || currentCluster.getNodes().length === 0">There are no nodes to show.</div>
-                    <div class="box col-md-2 nopadding" ng-repeat="node in currentCluster.getNodes()">
-                        <a class="text-center" ui-sref="nodeDetails({clusterName: currentCluster.name, nodeName: node.name})">
-              {{node.name}} <br><br>
-            </a>
+                <div ng-show="isNodesView()">
+                    <div ng-show="!currentClusterHasNodes()">There are no nodes to show.</div>
+                    <div class="box col-md-2 nopadding"
+                    ng-repeat="node in currentCluster.getNodes() | filter: {name: searchNameQuery}"
```
Martinsos: Unusal indentation

src/main/webapp/components/template/slider/slider.html
```diff
@@ -0,0 +1,3 @@
+<div id="slider"></div>
```
Martinsos: You should/must not use slider as id here: this is a reusable component, and id should not repeat. In reusable directive (reusable multiple times per page), you should not use any ids.

src/main/webapp/components/directives/slider/slider.js
```diff
@@ -0,0 +1,92 @@
+(function () {
+  'use strict';
+
+  var module = angular.module('ispn.directives.slider', []);
+
+  module.directive('slider', function () {
+      return {
+        restrict: 'E',
```
Martinsos: Use `replace` to remove custom elements from html

src/main/webapp/components/directives/slider/slider.js
```diff
@@ -0,0 +1,92 @@
...
+              return scope.$apply(function () {
+                scope.from = fromParsed;
+                scope.to = toParsed;
+              });
+            });
+
+            //watch model for changes
+            scope.$watch('from', function (newVal, oldVal) {
+              if (newVal !== fromParsed) {
+                return sliderElement.val([
```
Martinsos: why is return here, how are we returning this value to?

src/main/webapp/components/directives/slider/slider.js
```diff
@@ -0,0 +1,92 @@
...
+              toParsed = parseFloat(sliderElement.val()[1]);
+              return scope.$apply(function () {
+                scope.from = fromParsed;
+                scope.to = toParsed;
+              });
+            });
+
+            //watch model for changes
+            scope.$watch('from', function (newVal, oldVal) {
+              if (newVal !== fromParsed) {
```
Martinsos: Here I would rather use scope.from instead of fromParsed -> it seems more obvious what is happenning.

src/main/webapp/components/directives/slider/slider.js
```diff
@@ -0,0 +1,92 @@
...
+                max: [parseFloat(scope.max)]
+              }
+            });
+            sliderElement.on(sliderEvent, function () {
+              parsedValue = parseFloat(sliderElement.val());
+              return scope.$apply(function () {
+                scope.value = parsedValue;
+              });
+            });
+            return scope.$watch('value', function (newVal, oldVal) {
```
Martinsos: Same like before, there is no need here to compare with parsedValue, you can just update the slider element, or if you want to check just compare oldVal to newVal.

src/main/webapp/index.html
```diff
@@ -22,6 +26,8 @@
     <!-- build:js(src/main) scripts/modernizr.js -->
     <script src="bower_components/modernizr/modernizr.js"></script>
     <!-- endbuild -->
+
+    <link rel="stylesheet" href="components/directives/slider/jquery.nouislider.ispn.css" />
```
Martinsos: I would put this one just below line 23, so it will get built together with it.

src/main/webapp/login/login.controller.js
```diff
@@ -1,5 +1,29 @@
 'use strict';
 
+var ispn = ispn || {};
```
Martinsos: This is "wild" javascript code: it is global, and we should not write global code when using angular. So in this case we want to expose some util library, so it can be used by other components: you should make it a service.

Martinsos:
@vblagoje I wrote few small comments, other than that looks fine, good job with directive!
Pay special attention to implementing authentication rejection correctly, as it is important for both UX and handling users.

vblagoje:
Excellent stuff, much appreciated @Martinsos . I will make those adjustments today and then @tristantarrant can integrate this PR.

src/main/webapp/components/template/slider/slider.html
vblagoje: Riiight. I did not know how to grab it from document except to use id. I will work with some other API to grab the div component from the directive!

src/main/webapp/cluster-view/cluster-view.controller.js
vblagoje: Will do so! Good point!

src/main/webapp/cluster-view/cluster-view.controller.js
vblagoje: Excellent idea!

src/main/webapp/login/login.controller.js
vblagoje: Great, time to learn about Angular services :-)

src/main/webapp/cluster-view/cluster-view.controller.js
vblagoje: Yep, I see what you mean, more reuse! I'll do that.

src/main/webapp/cluster-view/cluster-view.controller.js
vblagoje: @Martinsos cool indeed and I tried it. However, it is not easy to work with ng-change on custom directive. I have to wire in the model and how to watch for changes on it. See example at https://docs.angularjs.org/api/ng/type/ngModel.NgModelController At this time this is way above my capabilities. Will revisit it in the future once I get a better understanding of javascript and Angular.

src/main/webapp/components/directives/slider/slider.js
vblagoje: I don't think it is needed as no ui slider works on div element. See http://refreshless.com/nouislider/examples/ If I set replace to true then slider disappears.

src/main/webapp/components/directives/slider/slider.js
vblagoje: True, not needed. I removed it

src/main/webapp/components/directives/slider/slider.js
vblagoje: I agree and I tried that as well. However, they way the slider [fires] (http://refreshless.com/nouislider/events-callbacks/) events makes the current approach more suitable. If I use the approach you propose (which is admittedly more concise and clear) the slider itself feels heavy and jittery.

src/main/webapp/index.html
vblagoje: +1 got it

vblagoje:
@tristantarrant I have addressed majority of @Martinsos concerns except the authentication rejection part. Please integrate this one @tristantarrant . We need to move on.

tristantarrant:
Great work, thanks

tsykora:
Thank you @vblagoje @Martinsos and @tristantarrant !!

That's one small step for a man, one giant leap for mankind!

src/main/webapp/components/template/slider/slider.html
Martinsos: Oh right, well the best way is just to use css class and grab it by that!

PR #10, opened 2015-04-23
Simplify gulp build

vblagoje:
Master only. Simplify dist build by eliminating tmp folder, avoiding injection, minification, template cacheing - anything complicated (for now). Files are served from dist build, changes are watched and acted upon, dist build could be zipped up and deployed anytime as a self contained unit.


vblagoje:
@Martinsos please have a look

vblagoje:
Although browsersync works for less/css file changes it does not for html and js. Not sure if it is possible to track these changes and reload without "breaking" the app. If we can fix that it would great!

src/main/gulp/e2e-tests.js
```diff
@@ -1,36 +0,0 @@
...
-      throw err;
-    })
-    .on('end', function () {
-      // Close browser sync server
-      browserSync.exit();
-      done();
-    });
-});
-
-gulp.task('protractor', ['serve:e2e', 'protractor-only']);
```
tsykora: Hmmmm, I suppose we need this file. I am running e2e protractor tests using: gulp protractor

src/main/gulp/unit-tests.js
```diff
@@ -1,31 +0,0 @@
-'use strict';
-
-var gulp = require('gulp');
-
-var $ = require('gulp-load-plugins')();
-
-var wiredep = require('wiredep');
-
-gulp.task('test', function() {
```
tsykora: Maybe same here? I would like to keep gulp.test task for Karma tests.

Martinsos:
@vblagoje I am very new with gulp, but it mostly seems ok to me! You removed some of optimizations, but as you said you can always add them later when you have better understanding of them.

I have a feeling that maybe custom defined pipes should not always do the write to files, but instead should output files through stream, and then in some task you can take all this pipes and write to where you want -> it makes them more reusable.

Also, I am not sure what is better:: combining all of the pipes into one task and then having build task that depends on clean and executes combined pipes task, or have each build task execute many tasks where those of them that depend on cleaning, have clean as their own dependency (I saw this on few places).

I am having trouble running an app, when I start infinispan server it breaks (I am using instructions from infinispan console README, and I am on master branch of infinispan (server):
```
13:47:36,925 ERROR [org.jboss.msc.service.fail] (MSC service thread 1-13) MSC000001: Failed to start service jboss.infinispan.clustered.namedCache: org.jboss.msc.service.StartException in service jboss.infinispan.clustered.namedCache: Failed to start service

```

src/main/gulp/e2e-tests.js
vblagoje: Ok I will put it back in. Where are these tests @tsykora ?

src/main/gulp/unit-tests.js
vblagoje: Ok

vblagoje:
@Martinsos at your convenience, please do git hard reset back to commit 869f778 as that one should work. It works for me. 

I like the pipe approach as pipes can be reused in tasks. If needed it is easy to convert pipe into the task by registering it with gulp - as we do at the bottom of build.js. This approach seems more modularized and allows fine control when creating tasks from these building blocks - pipes.  Notice how in watch.js we launch these gulp tasks for granular control of what gets updated/compiled/transferred to dist

Martinsos:
Ah yes I also like the pipes approach :)! What I wanted to suggest is that they could be even more modular if they would not contain `.pipe(gulp.dest(paths.somePath))`, instead you would add that depending on where you are using that pipe.

vblagoje:
Aha, I get it. Yeah, I'll see how that can be done! Thanks @Martinsos

Martinsos:
I am using this commit now: cd533dbd9f86ef4fec3ab9729e6688a9102ca882, and now I can start infinispan server, but I get this error after I login through web app:
```
[Server:server-one] ERROR [org.jboss.as.controller.management-operation] (ServerService Thread Pool -- 34) JBAS014612: Operation ("read-attribute") failed - address: ([
[Server:server-one]     ("subsystem" => "infinispan"),
[Server:server-one]     ("cache-container" => "security")
[Server:server-one] ]): java.lang.NullPointerException
```

vblagoje:
We need to handle better the case when we invoke a DMR op on a cluster container that's disabled. In this case security cache container. Everything else should work ok, I believe.

src/main/gulp/e2e-tests.js
tsykora: @vblagoje goje look here please: https://github.com/infinispan/infinispan-management-console/pull/4/files I will fix them to run with current master soon + re-issue fixed PR only with tests ;) and starting with development of others + external contributors should come up with more soon. You will know asap when we move somewhere here... don't worry.

vblagoje:
@Martinsos and @tsykora I believe I have addressed all your concerns. Have a look again, please.

Martinsos:
@vblagoje I wish you did not amend existing commit but instead just pushed another commit with fixes, because now it is hard to notice what you changed in the latest set of changes!

All is fine with me, nice job with simplifying the gulpfile and using the idea of reusable pipe segments!
I hope that when gulpfile grows we will extract destinations out from the pipe segments to make them more reusable. Also lets not forget that later it would be nice to add again wiredep (which we removed for now), so we do not have to worry about including all the files manually.

vblagoje:
@Martinsos I agree. Next time I will not amend but will continue adding commits. I agree about wiredep as well. Thank you for the review and your continued guidance. 

@tristantarrant this one is ready for integration. @tristantarrant you can simply issue command gulp build and build will be prepared in dist directory. See what else you need to do to disable proxy and wire the build into Infinispan server directly.

Martinsos:
@vblagoje no problem, I am glad to help when I can: I also learned a lot about gulp from you in this issue!

tristantarrant:
Pushed, thanks

PR #12, opened 2015-07-16
Adjust resource naming scheme to Infinispan 8.0

vblagoje:
Also included is a small gulp fix so that changes to JS and HTML while node server is running are immediately deployed. This part of the build did not work.


tristantarrant:
Pushed, thanks

PR #13, opened 2015-07-17
Create, edit, view cache and cache configuration

vblagoje:
@Martinsos please have a look! We should have a version of this PR integrated by Monday.

### DIFF
--- a/src/main/webapp/cache-details/cache-details.html
+++ b/src/main/webapp/cache-details/cache-details.html
@@ -1,17 +1,17 @@
 <div class="container-fluid" style="margin:20px">
     <div class="row">
-        <div class="col-md-6 breadcrumbs">
+        <div class="col-md-9 breadcrumbs">
             <span> <a href="#/cluster/{{currentCluster.name}}"> {{currentCluster.name}} </a>  >>  {{currentCache.name}} </span>
         </div>
 
-        <div class="col-md-6 text-right">
+        <div class="col-md-3">
             <div class="dropdown dropdown-menu-right">
                 <button class="btn btn-default dropdown-toggle" type="button" id="cacheActionMenu" data-toggle="dropdown">
                     Actions
                     <span class="caret"></span>
                 </button>
                 <ul class="dropdown-menu" role="menu" aria-labelledby="dropdownMenu1">
-                    <li role="presentation"><a role="menuitem" tabindex="-1" href="#">Edit cache</a>
+                    <li role="presentation"><a role="menuitem" tabindex="-1" ui-sref="editCache({clusterName: currentCluster.name, cacheName: currentCache.name})">Edit cache</a>
                     </li>
                     <li role="presentation"><a role="menuitem" tabindex="-1" href="#">Stop cache</a>
                     </li>
@@ -71,7 +71,7 @@
                 </div>
             </div>
             <div class="col-md-3 text-right">
-                <span class="search-text"> Search: </span> 
+                <span class="search-text"> Search: </span>
                 <input type="search" class="search-field">
             </div>
         </div>
@@ -83,63 +83,51 @@
                         <th>Name</th>
                         <th>Average reads</th>
                         <th>Average writes</th>
+                        <th>Average replication time</th>
                         <th>Total entries</th>
                         <th>Hits</th>
                         <th>Misses</th>
-                        <th>Read/Write ratio</th>
-                        <th>Total failed writes</th>
                     </thead>
                     <tr  ng-repeat="stats in currentCacheStats.nodeStats">
                         <td>{{stats['node-name']}}</td>
                         <td>{{stats['average-read-time']}} ms</td>
                         <td>{{stats['average-write-time']}} ms</td>
+                        <td>{{stats['average-replication-time']}}</td>
                         <td>{{stats['number-of-entries']}}</td>
                         <td>{{stats['hits']}}</td>
                         <td>{{stats['misses']}}</td>
-                        <td>{{stats['read-write-ratio']}}</td>
-                        <td>N/A</td>
                     </tr>
                 </table>
             </div>
         </div>
 
         <br>
-        <div class="row">
+      <div class="row">
             <div class="col-md-12">
                 <div class="cache-status">
-                    <span class="cache-status-label">Properties</span>
+                    <span class="cache-status-label">Configuration</span>
                 </div>
             </div>
+      </div>
+      <div class="cache-properties row">
+        <div><p><p>
         </div>
-
-        <div class="cache-properties row">
-            <div class="col-md-3">
-                <div>Cache name:</div>
-                <div>Mode:</div>
-                <div>Associated cache/loader store:</div>
-
-                <div>Write-through/Write-behind:</div>
-                <div>Isolation level:</div>
-                <div>Eviction strategy:</div>
-
-                <div>Passivation:</div>
-                <div>Expiration configuration:</div>
-                <div>Cross-site replication:</div>
-            </div>
-            <div class="col-md-9">
-                <div>{{currentCache.name}}</div>
-                <div>{{currentCache.type}}</div>
-                <div>N/A</div>
-
-                <div>Mock</div>
-                <div>{{currentCache.locking.LOCKING.isolation}}</div>
-                <div>N/A</div>
-
-                <div>N/A</div>
-                <div>N/A</div>
-                <div>N/A</div>
-            </div>
+        <div ng-if="currentCache.isDistributed()">
+          <cache-configuration cache-type="distributed-cache" metadata="currentCluster.cachesMetadata"
+                               current-cache="currentCache" init-defaults="false" read-only="true"/>
         </div>
-
+        <div ng-if="currentCache.isReplicated()">
+          <cache-configuration cache-type="replicated-cache" metadata="currentCluster.cachesMetadata"
+                               current-cache="currentCache" init-defaults="false" read-only="true"/>
+        </div>
+        <div ng-if="currentCache.isInvalidation()">
+          <cache-configuration cache-type="invalidation-cache" metadata="currentCluster.cachesMetadata"
+                               current-cache="currentCache" init-defaults="false" read-only="true"/>
+        </div>
+        <div ng-if="currentCache.isLocal()">
+          <cache-configuration cache-type="local-cache" metadata="currentCluster.cachesMetadata"
+                               current-cache="currentCache" init-defaults="false" read-only="true"/>
+        </div>
+      </div>
     </div>
 </div>
--- a/src/main/webapp/cluster-view/cluster-view.controller.js
+++ b/src/main/webapp/cluster-view/cluster-view.controller.js
@@ -7,8 +7,9 @@ angular.module('managementConsole')
     '$state',
     '$q',
     'modelController',
+    'cacheCreateController',
     'utils',
-    function ($scope, $stateParams, $state, $q, modelController, utils) {
+    function ($scope, $stateParams, $state, $q, modelController, cacheCreateController, utils) {
       if (!modelController.isAuthenticated()) {
         $state.go('/logout');
       }
--- a/src/main/webapp/components/api/cache-creation-controller.service.js
+++ b/src/main/webapp/components/api/cache-creation-controller.service.js
@@ -1,0 +1,1098 @@
+'use strict';
+
+angular.module('managementConsole.api')
+    .factory('cacheCreateController', [
+    '$http',
+    '$q',
+    'modelController',
+    function ($http, $q, modelController) {
+
+            /**
+             * Represents a client to the cacheCreateController
+             *
+             * @constructor
+             * @param {string} url - the URL to the ModelController management endpoint
+             */
+            var CacheCreationControllerClient = function (url) {
+                this.url = url + '/management';
+            };
+
+
+            /**
+             *
+             * Executes an operation
+             *
+             * @param data
+             * @param callback
+             */
+            CacheCreationControllerClient.prototype.execute = function (op) {
+                var deferred = $q.defer();
+                var http = new XMLHttpRequest();
+                var username = modelController.credentials.username;
+                var password = modelController.credentials.password;
+                http.withCredentials = true;
+                http.open('POST', this.url, true, username, password);
+                http.setRequestHeader('Content-type', 'application/json');
+                http.setRequestHeader('Accept', 'application/json');
+                http.onreadystatechange = function () {
+                    if (http.readyState === 4 && http.status === 200) {
+                        var response = JSON.parse(http.responseText);
+                        if (response.outcome === 'success') {
+                            deferred.resolve(response.result);
+                        } else {
+                            deferred.reject();
+                        }
+                    }
+                };
+                //console.log(JSON.stringify(op));
+                http.send(JSON.stringify(op));
+                return deferred.promise;
+            };
+
+
+            /**
+             *
+             * Creates a cache at a given DMR address a configuration structure cacheConfiguration
+             *
+             * @param address DMR address of the cache
+             * @param cacheConfiguration configuration
+             * @param callback to execute at the end of create operation
+             */
+            CacheCreationControllerClient.prototype.createCache = function (address, cacheConfiguration, callback) {
+              if (cacheConfiguration.type === 'distributed-cache' || cacheConfiguration.type === 'replicated-cache') {
+                this.createReplicatedOrDistributedCache(cacheConfiguration, address, cacheConfiguration.type, callback);
+              } else if (cacheConfiguration.type === 'invalidation-cache' || cacheConfiguration.type === 'local-cache') {
+                this.createInvalidationOrLocalCache(cacheConfiguration, address, cacheConfiguration.type, callback);
+              }
+            };
+
+
+            /**
+             *
+             * Creates a replicated/distributed cache at a given DMR address a and a cache configuration structure
+             *
+             * @param cache cache configuration
+             * @param address DMR address of the cache
+             * @param cacheType cache type to create
+             * @param callback to execute at the end of create operation
+             */
+            CacheCreationControllerClient.prototype.createReplicatedOrDistributedCache = function (cache, address, cacheType, callback) {
+              var promise = null;
+              if (cacheType === 'distributed-cache') {
+                promise = this.createDistributedCache(address, cache.data);
+              } else if (cacheType === 'replicated-cache') {
+                promise = this.createReplicatedCache(address, cache.data);
+              }
+              promise.then(function () {
+                this.createLockingNode(address.concat('locking', 'LOCKING'), cache.data.locking.LOCKING);
+                this.createEvictionNode(address.concat('eviction', 'EVICTION'), cache.data.eviction.EVICTION);
+                this.createExpirationNode(address.concat('expiration', 'EXPIRATION'), cache.data.expiration.EXPIRATION);
+                this.createCompatibilityNode(address.concat('compatibility', 'COMPATIBILITY'), cache.data.compatibility.COMPATIBILITY);
+                this.createTransactionNode(address.concat('transaction', 'TRANSACTION'), cache.data.transaction.TRANSACTION);
+                this.createStateTransferNode(address.concat('state-transfer', 'STATE_TRANSFER'), cache.data['state-transfer'].STATE_TRANSFER);
+                this.createFileStoreNode(address.concat('file-store', 'FILE_STORE'), cache.data['file-store'].FILE_STORE);
+                this.createJDBCStoreNode(address.concat('string-keyed-jdbc-store', 'STRING_KEYED_JDBC_STORE'),
+                  cache.data['string-keyed-jdbc-store'].STRING_KEYED_JDBC_STORE);
+              }.bind(this)).then(function () {
+                callback();
+              })
+            };
+
+            /**
+             *
+             * Creates a invalidation/local cache at a given DMR address a and a cache configuration structure
+             *
+             * @param cache cache configuration
+             * @param address DMR address of the cache
+             * @param cacheType cache type to create
+             * @param callback to execute at the end of create operation
+             */
+            CacheCreationControllerClient.prototype.createInvalidationOrLocalCache = function (cache, address, cacheType, callback) {
+              var promise = null;
+              if (cacheType === 'invalidation-cache') {
+                promise = this.createInvalidationCache(address, cache.data);
+              } else if (cacheType === 'local-cache') {
+                promise = this.createLocalCache(address, cache.data);
+              }
+              promise.then(function () {
+                this.createLockingNode(address.concat('locking', 'LOCKING'), cache.data.locking.LOCKING);
+                this.createEvictionNode(address.concat('eviction', 'EVICTION'), cache.data.eviction.EVICTION);
+                this.createExpirationNode(address.concat('expiration', 'EXPIRATION'), cache.data.expiration.EXPIRATION);
+                this.createCompatibilityNode(address.concat('compatibility', 'COMPATIBILITY'), cache.data.compatibility.COMPATIBILITY);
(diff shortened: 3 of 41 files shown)


vblagoje:
Here is the preview of this feature http://screencast-o-matic.com/watch/coibIyftHg

Martinsos:
Thanks @vblagoje, I will take a look at it today/tomorrow!

src/main/webapp/edit-cache/edit-cache.controller.js
```diff
@@ -0,0 +1,34 @@
...
+  .controller('editCacheCtrl', [
+    '$scope',
+    '$state',
+    '$stateParams',
+    'utils',
+    'modelController',
+    function ($scope, $state, $stateParams, utils, modelController) {
+      //-- Variables --//
+
+      if (!modelController.isAuthenticated()) {
```
Martinsos: While this also works, it would be better to check for authentication before entering controller, ideally on route change start! I think this should not be solved in this PR, but I suggest creating an issue for this, so it can be done later.

src/main/webapp/create-cache/create-cache.controller.js
```diff
@@ -0,0 +1,196 @@
...
+      }
+
+      var server = modelController.getServer();
+      var clusters = server.getClusters();
+      $scope.currentCluster = server.getCluster(clusters, $stateParams.clusterName);
+      $scope.newlyCreatedCache = {};
+      $scope.selectedCacheType = 'distributed-cache';
+      $scope.isCreateEnabled = false;
+
+      $scope.createReplicatedOrDistributedCache = function (cache, address, cacheType) {
```
Martinsos: This is a large function with cache-specific logic, so I would suggest moving it to some service that is related to cache or maybe a new service if non of existing services is good enough. It seems to me like there is some scope-related logic at the end of the function, but we can pass that as a callback to this function.

src/main/webapp/create-cache/create-cache.controller.js
```diff
@@ -0,0 +1,196 @@
...
+            cache.data['capacity-factor'], cache.data.batching, cache.data.start, cache.data['l1-lifespan'], cache.data['remote-site'],
+            cache.data['jndi-name'], cache.data['queue-size'], cache.data['queue-flush-interval'], cache.data.owners);
+        } else if (cacheType === 'replicated-cache') {
+          promise = cacheCreateController.createReplicatedCache(address, cache.data.mode, cache.data.module, cache.data['remote-cache'],
+            cache.data.indexing, cache.data['auto-config'], cache.data.statistics, cache.data['remote-timeout'],
+            cache.data.batching, cache.data.start, cache.data['remote-site'],
+            cache.data['jndi-name'], cache.data['queue-size'], cache.data['queue-flush-interval']);
+        }
+        promise.then(function () {
+          var lockingAddress = angular.copy(address);
```
Martinsos: There seems to some amount of repetition in this code, maybe it would be good to generalize it (with some function)? Or would generalazation here would be overengineering and reduce flexibility? Just giving an idea!

src/main/webapp/components/services/utils.js
```diff
@@ -56,6 +56,58 @@
...
+            o = o[n];
+          } else {
+            o[n] = {};
+            o = o[n];
+          }
+        }
+        o[a[a.length - 1]] = value;
+      },
+
+      getValue: function getValue(object, path) {
```
Martinsos: Some comments here would be useful :)!

src/main/webapp/components/directives/cache/security/security.js
```diff
@@ -0,0 +1,29 @@
+(function () {
```
Martinsos: I see that there is separate directive for each cache, that is because they are so different? It seems like a good approach, just asking!

src/main/webapp/components/directives/cache/replicated-cache.html
```diff
@@ -0,0 +1,47 @@
+<div>
+  <tabset>
+    <tab heading="General" active="true">
+      <generalreplicated data='currentCache' metadata="resolveDescription('general')" init-defaults="initDefaults" read-only="readOnly"/>
+    </tab>
+    <tab heading="Locking" active="false">
```
Martinsos: Just an idea: you might consider using ng-repeat here, where you would give ng-repeat list of objects with information for each tab! It may or may not be worth trouble to prepare this objects, I am not sure about what data is available and how is it formatted, but it would reduce some html repetition (and therefore also chances of mistake).
Same idea goes for tabs below!

src/main/webapp/components/directives/cache/remotestore/remotestore.js
```diff
@@ -0,0 +1,39 @@
...
+          initDefaults: '=',
+          readOnly:'='
+        },
+        replace: true,
+        templateUrl: 'components/directives/cache/remotestore/remotestore.html',
+        link: function (scope, element, attrs) {
+
+          if (scope.initDefaults){
+            scope.data = {};
+            scope.data.cache = scope.metadata.cache.default;
```
Martinsos: This could be done somewhat nicer with for loop!

src/main/webapp/components/directives/cache/remotestore/remotestore.html
```diff
@@ -0,0 +1,136 @@
+<div class="container-fluid">
+  <div class="row">
+    <div class="form-horizontal">
+      <div class="form-group">
```
Martinsos: Here I would certainly use ng-repeat for all this form groups! It will reduce repetition a lot.

src/main/webapp/components/directives/cache/local-cache.html
```diff
@@ -0,0 +1,43 @@
+<div>
```
Martinsos: ng-repeat also

Martinsos:
@vblagoje impressive work! I did a quick pass and left some comments, but there is a lot of files so I did not have time to go through it thorougly yet. Most of comments are related to using ng-repeat, and there are also many places where it would be awesome to use it which I did not comment, but I am sure you will recognize those on you own after you see those that I pointed out.
In directory cache/ there are many directives, could you shortly explain what are they for?
Even better, it would be great if you could add some more comments to the code, especially to explain what the directives do (header before each directive)!
I will take a more thorough look today/tomorrow, looking forward to your changes!

src/main/webapp/edit-cache/edit-cache.controller.js
vblagoje: Ok, acknowledged @Martinsos

src/main/webapp/create-cache/create-cache.controller.js
vblagoje: Agreed @Martinsos

src/main/webapp/create-cache/create-cache.controller.js
vblagoje: It is essentially all about taking input values from fields and invoking an API. These fields are likely not to change, but there is a slight possibility they might.

src/main/webapp/components/services/utils.js
vblagoje: I wish I could understand it myself. I borrowed it from SO :-) Will try to though.

src/main/webapp/components/directives/cache/security/security.js
vblagoje: Yes, for each cache configuration submodule there is an accompanying directive.

src/main/webapp/components/directives/cache/replicated-cache.html
vblagoje: Ok, so I would prepare those objects in Javascript part fo the directive?

src/main/webapp/components/directives/cache/remotestore/remotestore.js
vblagoje: Yes, but each directive has specific fields that need to be initialized. Service would go through field parameter names and initalize fields? Sounds interesting if it could be done.

vblagoje:
Thanks @Martinsos much appreciated!

src/main/webapp/create-cache/create-cache.controller.js
```diff
@@ -0,0 +1,196 @@
...
+          lockingAddress.push('locking', 'LOCKING');
+          var lockingNode = cache.data.locking.LOCKING;
+          cacheCreateController.createLockingChildNode(lockingAddress, lockingNode['acquire-timeout'], lockingNode.isolation,
+            lockingNode.striping, lockingNode['concurrency-level'])
+            .then(function () {
+              var evictionAddress = angular.copy(address);
+              var evictionNode = cache.data.eviction.EVICTION;
```
vblagoje: use concat instead of copy+push

src/main/webapp/create-cache/create-cache.controller.js
```diff
@@ -0,0 +1,196 @@
...
+        if (cacheType === 'invalidation-cache') {
+          promise = cacheCreateController.createInvalidationCache(address, cache.data.mode, cache.data.module, cache.data['remote-cache'],
+            cache.data.indexing, cache.data['auto-config'], cache.data.statistics, cache.data['remote-timeout'], cache.data.batching,
+            cache.data.start, cache.data['remote-site'], cache.data['jndi-name'], cache.data['queue-size'], cache.data['queue-flush-interval']);
+        } else if (cacheType === 'local-cache') {
+          promise = cacheCreateController.createLocalCache(address, cache.data.module, cache.data['remote-cache'],
+            cache.data.indexing, cache.data['auto-config'], cache.data.statistics, cache.data.batching, cache.data.start, cache.data['remote-site'],
+            cache.data['jndi-name']);
+        }
+        promise.then(function () {
```
vblagoje: no need to do invocation sequentially

src/main/webapp/components/directives/cache/configurationviewer.js
```diff
@@ -0,0 +1,53 @@
+(function () {
```
Martinsos: I suggest naming file the same like directive!

src/main/webapp/cache-details/cache-details.html
```diff
@@ -83,63 +83,51 @@
...
-
-                <div>Mock</div>
-                <div>{{currentCache.locking.LOCKING.isolation}}</div>
-                <div>N/A</div>
-
-                <div>N/A</div>
-                <div>N/A</div>
-                <div>N/A</div>
-            </div>
+        <div ng-if="currentCache.isDistributed()">
```
Martinsos: This ng-if part appears on multiple places through app, so we could eliminate it by moving this if-part to the cache-configuration directive. End result: we would just call `<cache-configuration cache-type="distributed-cache" ...></cache-configuration>`, and directive would inside itself take care that correct html is displayed. I will comment more about that in the directive code.

src/main/webapp/cache-details/cache-details.html
```diff
@@ -83,63 +83,51 @@
...
+          <cache-configuration cache-type="distributed-cache" metadata="currentCluster.cachesMetadata"
+                               current-cache="currentCache" init-defaults="false" read-only="true"/>
         </div>
-
+        <div ng-if="currentCache.isReplicated()">
+          <cache-configuration cache-type="replicated-cache" metadata="currentCluster.cachesMetadata"
```
Martinsos: Watch out, self closing angular directives are not allowed! It would be awesome if they were, but unfortunately angular has problem with those due to the way how browsers handle the self-closing tags: https://github.com/angular/angular.js/issues/1953

src/main/webapp/components/directives/cache/configurationviewer.js
```diff
@@ -0,0 +1,53 @@
...
+        scope: {
+          currentCache: '=',
+          metadata: '=',
+          initDefaults: '=',
+          readOnly: '=',
+          cacheType: '@'
+        },
+        replace: false,
+        templateUrl: function(elem, attrs) {
+          return 'components/directives/cache/'.concat(attrs.cacheType).concat('.html');
```
Martinsos: Instead of having multiple html files, one for each cache type, I would consider creating only one file (and it will be template of this directive) that will have ng-ifs in itself, to choose which tabs are shown depending on the type of cache.

src/main/webapp/cluster-view/cluster-view.controller.js
```diff
@@ -3,12 +3,14 @@
...
     '$scope',
+    '$modal',
     '$stateParams',
     '$state',
     '$q',
     'modelController',
+    'cacheCreateController',
     'utils',
-    function ($scope, $stateParams, $state, $q, modelController, utils) {
+    function ($scope, $modal, $stateParams, $state, $q, modelController, cacheCreateController, utils) {
```
Martinsos: I believe $modal is not used

src/main/webapp/components/api/cache-creation-controller.service.js
```diff
@@ -0,0 +1,1146 @@
...
+                this.url = url + '/management';
+            };
+
+
+            /**
+             * Executes an operation
+             * @param data
+             * @param callback
+             */
+            CacheCreationControllerClient.prototype.execute = function (op) {
```
Martinsos: Why are we not using $http service here? Is it not flexible enough?

src/main/webapp/components/api/cache-creation-controller.service.js
```diff
@@ -0,0 +1,1146 @@
...
+             * @param jndi-name	STRING	The jndi-name to which to bind this cache instance.
+             * @param queue-size	INT	In ASYNC mode, this attribute can be used to trigger flushing of the queue when it reaches a specific threshold.
+             * @param async-marshalling	BOOLEAN	If enabled, this will cause marshalling of entries to be performed asynchronously.
+             * @param queue-flush-interval	LONG	In ASYNC mode, this attribute controls how often the asynchronous thread used to flush the
+             * replication queue runs. This should be a positive integer which represents thread wakeup time in milliseconds.
+             * @param owners	INT	Number of cluster-wide replicas for each cache entry.
+             *
+             *
+             */
+            CacheCreationControllerClient.prototype.createDistributedCache = function (address, mode, module_param, remote_cache, segments, indexing,
```
Martinsos: Take config object, not each param by itself

src/main/webapp/components/api/cache-creation-controller.service.js
```diff
@@ -0,0 +1,1146 @@
...
+                                                                    auto_config, statistics, remote_timeout, capacity_factor,
+                                                                    batching, start, l1_lifespan, remote_site, jndi_name,
+                                                                    queue_size, queue_flush_interval, owners) {
+
+
+              var op = {
```
Martinsos: You can either just pass the given config object, or better and more secure: create new one by copying values from it. To copy values you can use for loop.

src/main/webapp/components/api/domain-model.service.js
```diff
@@ -56,7 +56,7 @@ angular.module('managementConsole.api')
                 promises.push(profilePromise);
 
                 this.servers = [];
-                var hostsPromise = this.modelController.readChildrenResources(this.getResourcePath(), 'host', 2, true, false).then(function (response) {
+                var hostsPromise = this.modelController.readChildrenResources(this.getResourcePath(), 'host', 2 , true, false).then(function (response) {
```
Martinsos: this space was added by accident I think!

src/main/webapp/components/directives/cache/configurationviewer.js
```diff
@@ -0,0 +1,53 @@
...
+        link: function (scope, element, attrs) {
+
+          if (!scope.readOnly) {
+            scope.readOnly = false;
+          }
+          if (!scope.initDefaults){
+            scope.initDefaults = false;
+          }
+
+          scope.resourceDescriptionMap = {};
```
Martinsos: If resourceDescriptionMap is used only inside this function (it is "private"), do not add it to scope, but make it local, like `var resourceDescriptionMap = {};`, not `scope.resourceDescriptionMap = {};`

src/main/webapp/components/directives/cache/configurationviewer.js
```diff
@@ -0,0 +1,53 @@
...
+      };
+    }
+  ]).directive("positiveInteger", function() {
+    return {
+      restrict: "A",
+
+      require: "ngModel",
+
+      link: function(scope, element, attributes, ngModel) {
+        ngModel.$validators.positiveInteger = function(str) {
```
Martinsos: I am hairsplitting here, but this is actually test for non-negative integer (>=), not for positive (>0): http://math.stackexchange.com/questions/18464/is-positive-the-same-as-non-negative

src/main/webapp/components/directives/cache/general/general.js
```diff
@@ -0,0 +1,37 @@
...
+        },
+        replace: true,
+        templateUrl: 'components/directives/cache/general/general.html',
+        link: function (scope, element, attrs) {
+          scope.cacheTypes = ['distributed-cache', 'local-cache','replicated-cache', 'invalidation-cache'];
+          scope.cacheModes = [ 'SYNC', 'ASYNC'];
+
+          if (scope.initDefaults){
+            scope.data.type = 'distributed-cache';
+            scope.data.data = {};
```
Martinsos: When creating new object, I think it is somewhat more intuitive/clearer to do it like this:
```
a = {
  b: 0
};
```
then like this
```
a = {};
a.b = 0;
```

src/main/webapp/create-cache/create-cache.controller.js
```diff
@@ -0,0 +1,196 @@
...
+          address.push($scope.newlyCreatedCache.type);
+          address.push($scope.newlyCreatedCache.name);
+          if ($scope.newlyCreatedCache.type === 'distributed-cache' || $scope.newlyCreatedCache.type === 'replicated-cache') {
+            $scope.createReplicatedOrDistributedCache($scope.newlyCreatedCache, address, $scope.newlyCreatedCache.type);
+          } else if ($scope.newlyCreatedCache.type === 'invalidation-cache' || $scope.newlyCreatedCache.type === 'local-cache') {
+            $scope.createInvalidationOrLocalCache($scope.newlyCreatedCache, address, $scope.newlyCreatedCache.type);
+          }
+      };
+
+      $scope.$on('createEnabled', function(e, data) {
```
Martinsos: Why are you using broadcasting? Could we do this communication by passing parameters to directives? That is more explicit.

src/main/webapp/create-cache/create-cache.controller.js
```diff
@@ -0,0 +1,196 @@
...
+        $scope.isCreateEnabled = data;
+      });
+
+      $scope.$on('createCacheTypeSelected', function(e, data) {
```
Martinsos: Same as above, why broadcasting? I would prefer passing attributes to directives if possible.

src/main/webapp/components/directives/cache/general/generaldist.js
```diff
@@ -0,0 +1,49 @@
...
+
+            scope.data.data.batching = scope.metadata.batching.default;
+            scope.data.data.module = scope.metadata.module.default;
+            scope.data.data.owners = scope.metadata.owners.default;
+            scope.data.data.segments = scope.metadata.segments.default;
+            scope.data.data.start = scope.metadata.start.default;
+            scope.data.data.statistics = scope.metadata.statistics.default;
+          }
+
+          scope.$watch("form.cacheName.$valid", function (){
```
Martinsos: What if cache name becomes valid first, but then it becomes invalid? Is it still ok to have create enabled? You could add broadcast of disable on invalid name, but the best option would be not to send messages through broadcasting mechanisms, but use directive attributes to share if creation is enabled or not.

src/main/webapp/components/directives/cache/remotestore/remotestore.js
Martinsos: Yes, you can do it!

src/main/webapp/components/services/utils.js
```diff
@@ -56,6 +56,58 @@
           return false;
         }
         return true;
+      },
+
+      setValue: function setValue(object, path, value) {
```
Martinsos: I like to name this type of function `deepSet`.
In javascript, if you have object `a = {b: 3}` and do `a.c.d = 2` you will get `can not set property d of undefined` error because c does not exist. What you want sometimes is for that `c` to be automatically created as an empty object.
That is why we use `deepSet` function (`setValue` here), that creates missing objects on the way, so you can say `deepSet(a, 'b.c.d.e', 5)` and it will do this: `a.b.c.d.e = 5` while creating needed objects if missing.

src/main/webapp/components/services/utils.js
Martinsos: Check out above, similar like setValue, but it gets value, and will not create any empty objects -> it will return undefined instead (instead of an error).

src/main/webapp/components/services/utils.js
```diff
@@ -56,6 +56,58 @@
...
+          if (n in o) {
+            o = o[n];
+          } else {
+            return;
+          }
+        }
+        return o;
+      },
+
+      makeResourceDescriptionMap: function makeResourceDescriptionMap(map){
```
Martinsos: Why are you receiving object and then filling it? Why not just create on instance of this object (singleton) and share it around? Is it going to be modified?

Martinsos: I am not sure if this function should go to utils. I usually prefer to keep only non-domain related functions in utils, but if you choose to keep those also then it is ok.

src/main/webapp/components/services/utils.js
```diff
@@ -56,6 +56,58 @@
...
+        map['tx'] = 'children.transaction.model-description.TRANSACTION.attributes';
+        map['statetransfer'] = 'children.state-transfer.model-description.STATE_TRANSFER.attributes';
+        map['filestore'] = 'children.file-store.model-description.*.attributes';
+        map['remotestore'] = 'children.remote-store.model-description.*.attributes';
+        map['jdbcstore'] = 'children.string-keyed-jdbc-store.model-description.*.attributes';
+        map['authorization'] = 'children.security.model-description.*.children.authorization.model-description.*.attributes';
+      },
+
+      resolveDescription: function resolveDescription (metadata, resourceDescriptionMap, elementPath, cacheType) {
+          var path = 'children.'.concat(cacheType).concat('.model-description.*');
```
Martinsos: You can just use `'children.' + cacheType + '.model-description'`

src/main/webapp/create-cache/create-cache.controller.js
Martinsos: We can get each of these to one or two lines I hope!

src/main/webapp/edit-cache/edit-cache.controller.js
```diff
@@ -0,0 +1,34 @@
...
+      var server = modelController.getServer();
+      var clusters = server.getClusters();
+      $scope.currentCluster = server.getCluster(clusters, $stateParams.clusterName);
+      $scope.caches = $scope.currentCluster.getCaches();
+      $scope.currentCache = $scope.caches[$stateParams.cacheName];
+      $scope.currentCache.refresh();
+
+
+      $scope.saveCache = function (){
+          var address = ['profile', 'clustered', 'subsystem', 'infinispan', 'cache-container', $scope.currentCluster.name];
```
Martinsos: Why not add those two immediately to array? Like this:
```
var address = ['profile', 'clustered', 'subsystem', 'infinispan', 'cache-container', $scope.currentCluster.name, $scope.newlyCreatedCache.type, $scope.newlyCreatedCache.name];
```

Martinsos:
@vblagoje I did the whole pass and added some more comments! I like it a lot all together, with few of these fixes we can probably make it somewhat more compact.

src/main/webapp/components/services/utils.js
vblagoje: Agreed! Thanks these pointers @Martinsos

src/main/webapp/components/directives/cache/remotestore/remotestore.html
vblagoje: I made this change @Martinsos This is fantastic!

tristantarrant:
Pushed

PR #16, opened 2015-09-07
Patternfly2.2

tristantarrant:
Upgrade to PatternFly 2.2


vblagoje:
@tristantarrant I tried it, looks very nice! However, I noticed this flicker of another screen during login process? Is that intentional?

vblagoje:
Integrated! Thanks @tristantarrant

PR #19, opened 2015-10-18
Cumulative updates up until ISPN-5760 

vblagoje:
Master only. I have bypassed editing of cache configuration when new cache is created (for the time being). Cache configuration editing will be added to the cache creation step very soon. One can view configuration of a cache, even now. Also, server node adding/starting works as well. @tristantarrant please review


tristantarrant:
Good progress @vblagoje pushing

PR #22, opened 2015-11-06
Cumulative updates (ISPN-5766 and ISPN-5764)

vblagoje:
Master only. @tristantarrant please review.


tristantarrant:
Good job !
A few notes:
* make sure that you match the height of all cards in the cache container and cluster view, including the "add" cards
* the mockup has a "styles.css" for collecting some console-specific styles, including variations to the PF styles. In particular .header-row { margin-bottom: 20px; } which creates some extra breathing space
* some of the cache traits (bounded and remote backup) use a different font because the text is inside the span that puts the icon.

vblagoje:
Ok @tristantarrant will take care of these over the weekend! Thanks

vblagoje:
I did not know how to address "some of the cache traits (bounded and remote backup) use a different font because the text is inside the span that puts the icon". If you know how, please do. Otherwise, we can do it later @tristantarrant

tristantarrant:
Good, one last thing: remove the status/events text. Those are mockups and until we can put stuff there, it's best to leave it blank

tristantarrant:
Pushed, thanks !

PR #24, opened 2015-11-20
ISPN-5971 Truncate long cache names

tristantarrant:
- Ellipsize the cache card names
- Fix padding between container status label and name
- Use default cursor on trait icons

### DIFF
--- a/src/main/webapp/cluster-view/cluster-view.less
+++ b/src/main/webapp/cluster-view/cluster-view.less
@@ -6,21 +6,10 @@
   margin-bottom: 0.5em;
 }
 
-.header-row {
-  margin-bottom: 20px;
-}
-
 .sidebar-pf fieldset label {
   display: block;
 }
 
-.col-md-2 a {
-  width: 100%;
-  color: inherit;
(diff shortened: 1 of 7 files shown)


src/main/webapp/cluster-view/cluster-view.less
```diff
@@ -29,18 +18,3 @@
   float: right;
   border: 1px solid #D8D8D8;
 }
-.card-link {
```
vblagoje: @tristantarrant this was a request from Rome f2f (click anywhere on the card to activate link)

vblagoje:
@tristantarrant I could not get ng-click to work on the card itself. I suggest we keep the current CSS that enables this functionality. If we find a way to do it somehow differently we can always amend the current solution. LMK

tristantarrant:
@vblagoje I have fixed it by wrapping the card divs in a's. I've also removed the version from the login page

vblagoje:
Integrated 0f030e36eda4874f55f4b677578c5205e833eaee Thanks @tristantarrant

PR #28, opened 2015-11-26
ISPN-5986 Avoid Nexus staging during the release

slaskawi:
This change allows to skip Nexus Staging Process.

Unfortunately the first test round will be on the release date. So far I've tested it using custom Nexus instance and I also tested snapshot releases (they are not affected).

https://issues.jboss.org/browse/ISPN-5986


tristantarrant:
Merged, thanks

PR #29, opened 2015-12-04
Admin console 8.1.0 final update

vblagoje:
Cummulative updates for 8.1.0 release. @tristantarrant please review
To try out indexing properties you need to enter JSON key/value pairs in the conf field. For example, you can use:
```
{"person": 50,"person2": 50}
```
For JDBC loader admin also needs to enter specific JSON fields in the table value. For example, in string-keyed-table use:
```
{"batch-size": 50,"id-column": {"name": "asdd","type": "sfsf"},"create-on-start": true,"data-column": {"name": "asdad","type": "sfsdf"}}
```

For security roles admin should use arrays. For example:
```
["admin", "new_admin"]
```

Let me know how it goes so I can address any issues ASAP.


vblagoje:
@tristantarrant note that in 1df2984 I switched order of default actions in cases when action is destructive or has large consequences (stopping/starting servers, starting/stoppping entire cluster and so on). Open to suggestions and corrective actions for course. This is very much gray area and I have little experience in it!

tristantarrant:
I've found a few issues:
- clicking on Clusters > Cluster > "Add Node" switches to an opaque login screen (i.e. with the click shield)
- clicking on Caches > "clustered" > Actions > View/Edit configuration goes back to the login screen

vblagoje:
Interesting @tristantarrant I can not reproduce the second. For me, nothing happens. I tried Firefox and Chrome. I'll put up on of those "Work in progress" dialogs for that action. I really don't know what you mean by the first issue you noticed. Please read through all the modal dialogs as well. Will work on further polishing tomorrow.

tristantarrant:
Ok, the 2nd issue is fixed, but the first one still happens:
I login, click on Clusters, click on main-server-group, click on Add Node
The screen changes back to the login screen, but I cannot do anything since the entire page is covered by the modal click shield, i.e. that opaque overlay that prevents you from clicking on anything but the modal dialog. However there is no dialog either.

tristantarrant:
Btw, I'm testing with infinispan master

vblagoje:
ok @tristantarrant let me try to reproduce the first issue. I'll report back soon. Anything else?

vblagoje:
@tristantarrant reproduce when I removed bower_components directory and did bower update. Must have been related to angular update we did. I'll investigate. @anistor you have to wait for me here please :-)

vblagoje:
@tristantarrant  @anistor Angular 1.4.4 works perfectly. And I have tested console with that Angular version extensively. I could not find a source of this error in 1.4.8, nothing in logs, debugged javascript console. Must be something related to inner details of 1.4.8 release. Since I do not see any major benefit of using 1.4.8 over 1.4.4 I rolled back to 1.4.4 which I know works well.

tristantarrant:
Merged, releasing 8.1.0.Final to maven now.
Thanks

vblagoje:
ok awesome @tristantarrant thank you

PR #36, opened 2016-01-19
Replace bower with npm

VaclavDedik:
Hi!

I know we talked about this last year, and you said you'd rather not replace bower with NPM. The problem is that we cannot productize management console with bower. I tried to find an alternative solution, but there isn't really anything. There are two major problems:

1. There is no shrinkwrap-like functionality for bower, i.e. you cannot lock down version of packages. There is an open issue in the bower repository that adds this feature, but even though a lot of people would like this feature, it hasn't been solved yet (for more than 2 years now). See https://github.com/bower/bower/issues/505
2. I couldn't find a way to use sinopia with bower. We need this so that it is possible to build the management console offline. It might be possible to use something else than sinopia with bower, but that would complicate the heck out of productization, so I am definitely against that.

So this commit is the best solution I came up with, unfortunately. It basically replaces bower with npm (which is already present anyway). There are two minor problems with this solution:

1. There is no package like "main-bower-files" that selects only the important files from the downloaded dependencies. The solution I implemented uses "gulp-npm-files" package (which selects runtime dependencies) and filter unnecessary files with gulp-filter.
2. Some packages might not be in npm repository while they might be in bower. For example, I had to ask c3-angular developer to publish his plugin in npm repository, but he did that so the package is currently in the npm repository (only in verison 1.0.1 though).

If you won't want to integrate this PR, it will be very hard for us to productize management console as we will have to integrate this commit in our product repository only, which will create many problems whenever we need to backport upstream changes.


vblagoje:
@VaclavDedik if your approach will enable us to continue to develop management console webapp seamlessly as we do now, and you will continue to maintain the bower replacement setup I am not totally against it. Thoughts @tristantarrant @pzapata

VaclavDedik:
Of course it will enable you to continue to develop the management console, basically the only thing that changes is that the dependencies has to be declared in package.json instead of bower.json. Sometimes, it might be necessary to modify the gulp build files, but generally that shouldn't be necessary.

tristantarrant:
Pushed, thanks @VaclavDedik

mgencur:
Hey, some people in my team now see this error when building the console with mvn clean package: https://gist.github.com/mgencur/4178016bbbced8f7f02e

The build requires running under "root". Could we add a flag for skipping the installation or solve this in another way? Thanks

tristantarrant:
@mgencur have you deleted node_modules and node directories ?

VaclavDedik:
@mgencur It's weird npm install is trying to resolve node_modules/.bin/bower when bower is no longer used. I get no such error. Isn't it possible that you (or some people in your team) forgot to clean the directory? (i.e. with "git clean -xdf")?

mgencur:
Looks like cleaning the directory helped me get further, but I'm still getting this one: https://gist.github.com/mgencur/2f35d1095db609e448c9

VaclavDedik:
Try "sudo chown -R $(whoami) ~/.npm", see http://stackoverflow.com/questions/16151018/npm-throws-error-without-sudo

mgencur:
Hmmmmm. Thanks @VaclavDedik  !! . Works now :)

vblagoje:
@VaclavDedik and @mgencur I am not able to build the project. I followed the instructions above. I erased my node_modules, I did "git clean -xdf" and "sudo chown -R $(whoami) ~/.npm" but I stil get https://gist.github.com/vblagoje/28caff9d8861fd8748bb I did mvn clean install when I got the the error in the gist. I also have the following node and npm versions on my mac osx mavericks:
beleza:infinispan-management-console vblagoje$ node --version
v4.2.1
beleza:infinispan-management-console vblagoje$ npm --version
3.3.8
beleza:infinispan-management-console vblagoje$ 

Any help appreciated!

VaclavDedik:
Looks like there is some gyp config file missing, I'd try to reinstall node-gyp, or the whole node.js.

vblagoje:
I was not able to resolve this issue. However, regular dev cycle (gulp clean serve) works for me. The issue I am experiencing is https://github.com/websockets/bufferutil/issues/25

vblagoje:
For me the most important things is that I can continue dev on management console. I hope this will get resolved sometimes soon so I can release projects as well - i.e. successfully execute 'mvn clean install'

vblagoje:
@tristantarrant resolved this one with https://github.com/infinispan/infinispan-management-console/commit/4e552b93339d1772ed2ef4099e9cf2881a4a3d48 Thanks a bunch

PR #38, opened 2016-01-28
ISPN-5814 View cross site replication state

pzapataf:
Please, note this PR depends on merging ISPN-6063 first.


vblagoje:
Nice @pzapataf Will take a look eod today! Cheers!

src/main/webapp/cache-status/cache-status.html
```diff
@@ -109,7 +109,33 @@ <h2 class="card-pf-title">
 
     <div class="row row-cards-pf">
       <!-- BEGIN CARD -->
-      <div class="col-xs-2 col-sm-2 col-md-2" match-height="ispn-card">
+      <div class="col-xs-6 col-sm-4 col-md-3" match-height="ispn-card"
+           ng-if="currentCacheStats.firstServerStats['offline-sites'].length > 0
```
vblagoje: @pzapataf make this one a method in controller. The logic in view should be kept to a minimum!

src/main/webapp/clusters-view/clusters-view.controller.js
```diff
@@ -12,9 +12,60 @@ angular.module('managementConsole')
...
+            // Updates the map of remote site status
+            //
+            $scope.refreshRemoteSitesStatus = function(cluster) {
+              var resourcePathCacheContainer = cluster.domain.getFirstServer().getResourcePath()
+                 .concat('subsystem', 'datagrid-infinispan', 'cache-container', cluster.name);
+
+                // Refresh list of offline sites
+                cluster.modelController.readAttribute(resourcePathCacheContainer, 'offline-sites').then(
+                  function (response){
+                      if( response != null && response.constructor === Array) {
```
vblagoje: No need for response != null check - I think. Because if this call succeeds then response has to be valid. Try it.

src/main/webapp/clusters-view/clusters-view.controller.js
```diff
@@ -12,9 +12,60 @@ angular.module('managementConsole')
...
+                        $scope.offlineSites[cluster.name] = response
+                      } else {
+                        $scope.offlineSites[cluster.name] = [];
```
vblagoje: Perhaps you can initialize all these values somewhere and then you would not need else branch

src/main/webapp/clusters-view/clusters-view.controller.js
```diff
@@ -12,9 +12,60 @@ angular.module('managementConsole')
...
+                // Refresh list of mixed sites
+                cluster.modelController.readAttribute(resourcePathCacheContainer, 'mixed-sites').then(
+                  function (response){
+                    if( response != null && response.constructor === Array) {
+                      $scope.mixedSites[cluster.name] = response
+                    } else {
+                      $scope.mixedSites[cluster.name] = [];
+                    }
+                  }
+               );
```
vblagoje: You can also add catch function to promise chain. I plan to do this throughout our code in the next pass so it is not a must. But see last screenshot on http://terussell85.github.io/blog/2013/12/23/the-angularjs-promise-anti-pattern-that-makes-me-cry/ and how it can be done.

src/main/webapp/cache-status/cache-status.html
```diff
@@ -109,7 +109,33 @@ <h2 class="card-pf-title">
...
+        <div class="card-pf card-pf-accented ispn-card">
+          <div class="card-pf-heading">
+            <h2 class="card-pf-title">Backup sites</h2>
+          </div>
+          <div class="card-pf-body">
+            <div class="row">
+              <div class="col-md-12">
+                <table class="col-sm-12 col-xs-12 col-md-12">
+                  <tbody>
+                  <tr ng-repeat="site in currentCacheStats.firstServerStats['offline-sites']"><th><span class="pficon pficon-error-circle-o"></span> <span class="label label-danger">{{site}}</span></th></tr>
```
vblagoje: You can also encapsulate currentCacheStats.firstServerStats['offline-sites'] and the likes under some function in scope. Perhaps it can be reused and we do not leak as much details from the controller into view. I mean, I am also learning about these details myself :-)

src/main/webapp/cache-status/cache-status.html
pruivo: as sugested by @tristantarrant, I've changed the names of the attributes:
https://github.com/infinispan/infinispan/pull/3948#discussion_r51123203

the current names are ```sites-online```, ```sites-offline``` and ```sites-mixed```

tristantarrant:
This has already been merged, closing

PR #40, opened 2016-02-04
ISPN-5813 View of cache container configuration 

vblagoje:
We implemented 4 out of 6 cache configuration tabs. But let's integrate this work now and more will follow in subsequent PRs. Please review @tristantarrant and @pzapataf Especially have a look at require restart mechanism and test implementations (threadpool and transport) as we will use it throughout codebase soon.

### DIFF
--- a/src/main/webapp/cache-container/cache-container-configuration-service.js
+++ b/src/main/webapp/cache-container/cache-container-configuration-service.js
@@ -81,12 +81,33 @@ angular.module('managementConsole')
         });
       }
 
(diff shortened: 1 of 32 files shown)


tristantarrant:
Since you can restart the cluster, why not hook up the corresponding actions in the container menu ?

vblagoje:
@tristantarrant I can, but it won't be on a container level, only on per server group level.

vblagoje:
Integrated with e7aba16d6e048cf3b4d6404d772f9c298267830f

PR #43, opened 2016-02-12
ISPN-6111 Management console - total number of entries is not displayed

vblagoje:
Master only. Resolves ISPN-6111 and improves logic of loading DMR data needed in node view


tristantarrant:
Pushed, thanks

PR #44, opened 2016-02-12
ISPN-6077 AddCache dialog only offers cache templates from clustered container

vblagoje:
Master only. Please see ISPN-6077 for more details.


tristantarrant:
Pushed, thanks

PR #45, opened 2016-02-12
ISPN-6081 Creating new cache with name that already exists leads to no response in the UI

vblagoje:
Master only. Careful with integration of this one as it contains other not yet integrated pull request https://github.com/infinispan/infinispan-management-console/pull/44


tristantarrant:
Pushed, thanks

PR #47, opened 2016-02-16
Small function ordering/definition fix

vblagoje:
Master only. Please try first as well @tristantarrant


vblagoje:
Integrated already. Closing.

PR #48, opened 2016-02-17
ISPN-5813 Initialize security and basic upload/remove

vblagoje:
Master only. @tristantarrant @pzapataf I got upload and remove to work for deployments but I could not get DMR op for deploy and undeploy correctly. The ones you found Tristan do not work, so I attempted variations, different endpoints to no luck. I downloaded Wildfly and tried to decode their dmr ops for deploy and undeploy - they seem not to use management-upload but rather management endpoint directly. However, they seem to use composite ops for deploy and undeploy. I need a bit more time to figure these out. I think it is ok if we release beta2 with this PR included. At least, deployments can be uploaded/removed - we'll get the rest done soon! Please try it out on a fresh server build without proxy before integrating. Thank you!

### DIFF
--- a/src/main/webapp/cache-container/cache-container-configuration-service.js
+++ b/src/main/webapp/cache-container/cache-container-configuration-service.js
@@ -59,26 +59,33 @@ angular.module('managementConsole')
       }
 
       /**
-       * TODO does not work yet
+       *
+       *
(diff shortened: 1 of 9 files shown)


tristantarrant:
Pushed, thanks

PR #50, opened 2016-02-18
Use relative paths in templates to fix the task dialog box

tristantarrant:


mgencur:
Integrated. Thanks @tristantarrant  for the quick fix.

PR #53, opened 2016-02-18
ISPN-6110 - memory values should be in GB not in bytes

vblagoje:
Master only. Use MB instead of bytes. I chose MB overGB because the values I've seen were rather small for GBs and issue reporter suggested we can use MB as well.


vblagoje:
@tristantarrant rebased to latest master

tristantarrant:
Pushed, thanks

PR #56, opened 2016-02-22
ISPN 5966 - user credentials are now persisted and user is automatica…

Martinsos:
…lly authenticated.
Now, when user reloads the page, he stays logged in and on it.
@vblagoje please check this out! I wrote 3 TODO's that are actually questions, I would like to hear your opinion on those.

### DIFF
--- a/src/main/webapp/components/api/api.js
+++ b/src/main/webapp/components/api/api.js
@@ -9,4 +9,5 @@ angular.module('managementConsole.api', [
   'ispn.directives.cache.cacheconfiguration',
   'ispn.directives.cache.configurationsection',
   'ispn.directives.cache.jdbcstore',
+  'LocalStorageModule'
 ]);
--- a/src/main/webapp/components/api/model-controller.service.js
+++ b/src/main/webapp/components/api/model-controller.service.js
@@ -7,7 +7,7 @@ angular.module('managementConsole.api')
     'DomainModel',
     'utils',
     'localStorageService',
-    function ($http, $q, DomainModel, utils, localStorageService) {
+      function ($http, $q, DomainModel, utils, localStorageService) {
             /**
              * Represents a client to the ModelController
              * @constructor
--- a/src/main/webapp/components/api/model-controller.service.js
+++ b/src/main/webapp/components/api/model-controller.service.js
@@ -6,7 +6,8 @@ angular.module('managementConsole.api')
     '$q',
     'DomainModel',
     'utils',
-    function ($http, $q, DomainModel, utils) {
+    'localStorageService',
+    function ($http, $q, DomainModel, utils, localStorageService) {
             /**
              * Represents a client to the ModelController
              * @constructor
@@ -17,8 +18,8 @@ angular.module('managementConsole.api')
                 this.url = url + '/management';
                 this.authenticated = false;
                 this.credentials = {
-                    username: null,
-                    password: null
+                    username: localStorageService.get('username') || null,
+                    password: localStorageService.get('password') || null
                 };
                 this.server = null;
             };
@@ -36,22 +37,27 @@ angular.module('managementConsole.api')
             };
 
             /**
-             * Logs into the management endpoint and determines the launch type
+             * Logs into the management endpoint and determines the launch type.
+             * If username and/or password are not given, previously stored username and/or password are used.
              * @param {string} username - the username to use when connecting to the management endpoint
              * @param {string} password - the password to use when connecting to the management endpoint
              */
             ModelControllerClient.prototype.login = function (username, password) {
-                this.credentials.username = username;
-                this.credentials.password = password;
+                if (username !== undefined) { this.credentials.username = username; }
+                if (password !== undefined) { this.credentials.password = password; }
                 return this.readResource([], false, true).then(function (response) {
+                // TODO: Handle situation where response is received but login failed - that happens!
                     this.authenticated = true;
+                    localStorageService.set('username', this.credentials.username);
+                    localStorageService.set('password', this.credentials.password);
                     var launchType = response['launch-type'];
                     if (launchType === 'DOMAIN') {
                         this.server = new DomainModel(this, response);
                     } else if (launchType === 'STANDALONE') {
                         //TODO
                     }
-                }.bind(this)).catch(function(e){
+                }.bind(this)).catch(function(e) {
+                  this.logout();
                   throw e;
                 });
             };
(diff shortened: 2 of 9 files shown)


src/main/webapp/components/api/model-controller.service.js
```diff
@@ -36,22 +37,27 @@ angular.module('managementConsole.api')
...
              * @param {string} username - the username to use when connecting to the management endpoint
              * @param {string} password - the password to use when connecting to the management endpoint
              */
             ModelControllerClient.prototype.login = function (username, password) {
-                this.credentials.username = username;
-                this.credentials.password = password;
+                if (username !== undefined) { this.credentials.username = username; }
+                if (password !== undefined) { this.credentials.password = password; }
                 return this.readResource([], false, true).then(function (response) {
+                // TODO: Handle situation where response is received but login failed - that happens!
```
vblagoje: When login fails we get an exception @Martinsos What I discovered is that we have to bind this in the catch block in order to properly logout - otherwise we never clean credentials from local storage

src/main/webapp/login/login.controller.js
```diff
@@ -22,6 +22,7 @@ angular.module('managementConsole')
         modelController.login(credentials.username, credentials.password).then(function () {
           $scope.authenticated = true;
           $scope.page.htmlClass = '';
+          // TODO: Why is this refreshing done here?
           var modelPromise = modelController.refresh();
```
vblagoje: I had to redo how the whole model is loaded on login @Martinsos We can review that part a bit later but I suggest we leave refresh call here for now

src/main/webapp/management-console.js
```diff
@@ -304,38 +305,60 @@ angular.module('managementConsole', [
...
+    '$rootScope', 'modelController', '$urlRouter', '$state',
+    function ($rootScope, modelController, $urlRouter, $state) {
+      $rootScope.$on('$stateChangeStart', function (event, toState) {
         // redirect to login page if not logged in
-        if ($location.path() !== '/login' && !modelController.isAuthenticated()) {
-          $location.path('/login');
+        if (toState.name !== 'login' && !modelController.isAuthenticated()) {
+          event.preventDefault();
+          modelController.login().then(function () {
+            // TODO: Do I have to do this refreshing here? Why?
```
vblagoje: So this is a call triggered on browser refresh? If so then we have already refreshed/loadded the model in login function called before this function in a promise chain. Therefore I think we do not need this call here. I'll try it myself

vblagoje:
@Martinsos I left the comments. Thank you so much - this fix saved the day for us!

src/main/webapp/components/api/api.js
```diff
@@ -9,4 +9,5 @@ angular.module('managementConsole.api', [
   'ispn.directives.cache.cacheconfiguration',
   'ispn.directives.cache.configurationsection',
   'ispn.directives.cache.jdbcstore',
+  'LocalStorageModule'
```
tristantarrant: Maybe we should use sessionStorage instead: https://www.owasp.org/index.php/HTML5_Security_Cheat_Sheet#Local_Storage

src/main/webapp/management-console.js
vblagoje: Hmm, I think we need call here because we do not use login controller to login. Admittedly I made things a bit more convoluted than they need to be @Martinsos

src/main/webapp/components/api/model-controller.service.js
Martinsos: But I am experiencing situation where I get response from the server although I have sent credentials that are both `undefined`! And `.then` block is executed.

src/main/webapp/management-console.js
Martinsos: It is ok actually - there are two ways a user can login, either by this part of code which executes on any state change (and also on (re)load of application), or through login controller.
If his credentials are stored and they are correct, login will normally happen here. If he logouts and credentials are erased (or he has never logged in before) this part of code will reject him and send him to login controller, where he will be able to do a login by inputing his credentials.

Model refresh will not happen twice because once that login executes successfuly (either here or in login controller) that is remembered in modelController (isAuthenticated() method) and since we check for that method here, it can not happen again!

I asked if we need this refresh here because I did not know what it is for. But if we needed it in login controller (as you said), then we need it here also.

src/main/webapp/components/api/api.js
Martinsos: @tristantarrant thanks, that sounds good! I investigated it a little bit and sessionStorage is localized per tab/window while localStorage is persistent, which makes sessionStorage a better choice for storing credentials. However, user will have to log in each time he opens a new tab/window - I guess that is ok in this case.
Also, we should keep in mind here that this is still not secure - for real security we should use ssl connection, tokens/session/cookies instead of credentials, use proper cors restrictions - I am not expert on this so there is certainly more things to watch out for, and in the future it would be good to study this thoroughly.

I changed storage type from localStorage to sessionStorage. I also confirmed that localStorageService has 'sessionStorage' set as its type, and that it is supported by my browser, however I could was not able to get the behaviour that I expected - when I opened infinispan console in another window, I was logged in. I am not sure, but I think it may be due to the BrowserSync syncing these tabs.

src/main/webapp/components/api/model-controller.service.js
Martinsos: Replicating the bug:

0. make sure that you are logged out of app
1. type http://localhost:3000/index.html#/clusters in url - app will open but with no name in upper right corner. That is because login got response, set authenticated to true but has no username really set. Here it should not get a response but fail instead, or it should be able to differentiate that response.
2. go to login screen, login - you willl see your username in upper right corner - that is how it should work.

Martinsos:
@vblagoje I replaced localStorage with sessionStorage (thanks @tristantarrant) - I elaborated in line comment.
One thing that we yet have to fix is problem with receiving response even when credentials are invalid - I see that github hid that comment but I wrote how to replicate the bug in line comment - check it under "outdated diff".

vblagoje:
@Martinsos Please see this fix [1] for the issue you described. This way we can prevent anyone to see any routes without authenticating first.

[1] https://github.com/vblagoje/infinispan-management-console/commit/d8022846b8611fb020b04167c551c3cf3956e6bb

Martinsos:
@vblagoje seems good! So the problem was in empty credentials? Is there any other way instead of using this randomly generated credentials - this seems like something that should be fixed on the server side, but I guess this is not the option?

vblagoje:
Not at this time @Martinsos What we have now is good enough and we can iterate further after next release. Thank you for this contribution once again!

vblagoje:
Integrated with cb99bc1ec42aa87dce6ae319d23aa27861ea00b9 Thanks @Martinsos

Martinsos:
@vblagoje understood, no problem!

PR #57, opened 2016-02-23
ISPN-6144 Display error messsages to user

vblagoje:
Master only. Do not *integrate* yet. For @tristantarrant and @pzapataf to try it out and test.


vblagoje:
Integrated with 6e150e20958ad0b97d3fe128b00ce15c50f88142

PR #59, opened 2016-02-24
Fix the cache containers breadcrumb when editing a cache configuration

tristantarrant:


vblagoje:
I also added a small breadcrumb fix and appended it to your commit @tristantarrant . Resolved with 856a43ccd239c75f2e477b892f1661fb8d9aa358

PR #62, opened 2016-02-29
ISPN-6284 Clicking refresh multiple times results in nodes being show…

vblagoje:
Master only. Check JIRA for issue description and test it yourself.


vblagoje:
Itegrated 679c85b395b8a596af9d9296731d6e8367cb129a

PR #63, opened 2016-03-01
ISPN-6282 - Script deployments

pzapataf:


tristantarrant:
Thanks @pzapataf . I managed to create / modify / delete / run scripts. I like the automatic creation of the parameter / type /name metadata.
Pushed

PR #64, opened 2016-03-02
ISPN-6117 Cache configuration can not be be edited

vblagoje:
Master only. *Do not integrate*, try it out @pzapataf and @tristantarrant . I will do the final polishing tomorrow.


vblagoje:
@tristantarrant and @pzapataf I added proposed solution to ISPN-6263 . Please have a look

src/main/webapp/management-console.js
```diff
@@ -5,8 +5,10 @@ angular.module('managementConsole', [
   'gridshore.c3js.chart',
   'ui.router',
   'ui.bootstrap',
-  'LocalStorageModule',
-]).directive('fileModel', ['$parse', function ($parse) {
+  'LocalStorageModule'
+]).constant('CONSTANTS', {
+  'NO_BASE_CONFIGURATION_TEMPLATE': 'NONE - no base template'
```
tristantarrant: I think it would be best to use the first empty option of that dropdown and make it the default instead of creating a new one for this case. I'd also either leave the text as blank or use angular (hohoho) brackets like "<none>"

tristantarrant:
A few notes:
- the global templates section in the cache container configuration should receive the same treatment
- I tried creating a cache from scratch, and it wanted me to set the name of the template. For the `<none>` use-case, the template-name should be pre-set to the cache name
- when I manually set the name to a different name, it shows "restart needed" on the right, but this should not happen a new template.
- for a new clustered cache the `mode` attribute is compulsory and the user needs to choose either SYNC or ASYNC. If this is too difficult to handle, just default it to SYNC.

src/main/webapp/management-console.js
vblagoje: If you do not mind let's use ```<none>``` as I already have check mechanisms in place to make sure that something is selected before proceeding to next screen. We can fine tune this for 8.3

vblagoje:
@tristantarrant all points are valid. The mode is the must to create a config. Another thing I can quickly add is to show all defaults so user has some reference point for values when creating a cache configuration. Super easy to do, and I believe it will help users a lot. Thoughts @tristantarrant and @pzapataf ?

tristantarrant:
Ok. In the tooltip ?

vblagoje:
Well, I thought to insert them in the fields directly @tristantarrant . We have them ready in metadata I just have to turn one switch on and they will be inserted in the model. Even if we do not show them to users they will be assigned anyways and this way users can see what they are, get a reference value and change them right there if they need to.

vblagoje:
@tristantarrant and @pzapataf I have addressed all Tristan's concern. I have added default values in placeholders when creating a new cache/template. Placeholders are not really possible for menu dropdown and checkbox. I have added default values to tooltips as well. I have tested all cache/template creation scenarios but I have not tested all configuration settings - plan to do that tonight. Please test this PR although it seems ready for me.

PR #66, opened 2016-03-03
ISPN-6291 Send REST queries to coordinator instead of first server cluster

vblagoje:
@pzapataf I need you to review this one as most of the changes I made are affecting the code you wrote. Please be thorough and check everything. I suggest we first integrate this PR and then https://github.com/infinispan/infinispan-management-console/pull/64


tristantarrant:
Pushed, thanks

PR #68, opened 2016-03-03
Better server group per cache container resolution

vblagoje:
Master only. Easier way to figure on which server group is cache container (ClusterModel) deployed on.


tristantarrant:
Looks good, pushed !

PR #70, opened 2016-03-04
ISPN-6229 Use "Task" insteade of "Type" in Launch Task dialog

tristantarrant:
https://issues.jboss.org/browse/ISPN-6229


vblagoje:
Resolved with d205abe

PR #72, opened 2016-03-15
Real boot dialogs, improved cluster page refresh

vblagoje:
Master only. Reworked page refresh on cluster view and node pages related to node starting, stopping, removing as well as adding new nodes. Using real boot dialogs. Please review @pzapataf @tristantarrant and *do not* integrate yet


pzapataf:
It works much better now. One small tweak I'd do is replace the empty memory usage  chart when a node is stopped by a message 'Node is stopped'.

vblagoje:
@pzapataf fixed the commit log ISPN-6140 reference and included empty memory screen. The message "Node is stopped" looked odd a bit and redundant as well as the red-colored "STOPPED" indicator is already indicating node status. I guess we can polish it further if needed.

vblagoje:
@pzapataf @tristantarrant I added another commit with Pedro's suggestion. Please choose which one you like better, squash and integrate

tristantarrant:
Squashed and pushed. Nice !

PR #73, opened 2016-03-16
ISPN-6230 Task history tab does not show tasks unless Refresh button is pressed as well

vblagoje:
Master only. Includes simplification for task handling and refresh fix as well. Please review @pzapataf @tristantarrant


tristantarrant:
Pushed, thanks

PR #74, opened 2016-03-18
ISPN-6339 Add mass reindex operation

vblagoje:
Master only. Eviscerate cache ops from cache model and move them to service + add reindex op only for indexable caches


src/main/webapp/cache-status/confirmation-cache-modal.html
```diff
@@ -10,6 +10,7 @@ <h4 class="modal-title" id="myModalLabel">Confirmation</h4>
     <h3 ng-if="mode === 'enable'">Enabling cache {{cacheName}} will attach it to all active remote endpoints. Enable {{cacheName}}?</h3>
     <h3 ng-if="mode === 'disable'">Disabling cache {{cacheName}} will detach it from all active remote endpoints. Disable {{cacheName}}?</h3>
     <h3 ng-if="mode === 'purge'">Purging cache {{cacheName}} will passivate all its cache entries. Purge {{cacheName}}?</h3>
+    <h3 ng-if="mode === 'reindex'">Reindex cache {{cacheName}} will index all its cache entries according to deinfed indexing settings. Reindex {{cacheName}}?</h3>
```
tristantarrant: Please surround all {{cacheName}} with quotes so that it is easier to see what is being acted upon.
Also I'd reword it to "Re-indexing cache '{{cacheName}}' will remove its current index and recreate it according to the cache contents' schema."

vblagoje:
Updated @tristantarrant

PR #75, opened 2016-03-22
Cummulative usability testing tweaks

vblagoje:
Various tweaks. @tristantarrant please review and adjust if needed.

### DIFF
--- a/src/main/webapp/cache-container/configuration-security/security.html
+++ b/src/main/webapp/cache-container/configuration-security/security.html
@@ -80,7 +80,6 @@
         <div class="form-group">
           <div class="col-md-10 col-md-offset-2">
             <button type="button" class="btn btn-primary" ng-click="defineAuthorization()">Yes</button>
-            <button type="button" class="btn btn-default">No</button>
           </div>
         </div>
       </div>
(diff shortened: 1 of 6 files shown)


src/main/webapp/cache-container/configuration-security/security.html
```diff
@@ -80,7 +80,6 @@
         <div class="form-group">
           <div class="col-md-10 col-md-offset-2">
             <button type="button" class="btn btn-primary" ng-click="defineAuthorization()">Yes</button>
```
tristantarrant: I prefer verbs on buttons. Therefore "Enable Authorization"

src/main/webapp/cluster-nodes/cluster-nodes.html
```diff
@@ -72,10 +72,11 @@ <h2 class="card-pf-title">Add Node
             inetAddress:server.getInetAddress()})" title="View Node Details">
           <div class="card-pf card-pf-accented card-pf-aggregate-status card-pf-with-action">
             <h2 class="card-pf-title">
-              <span class="pficon" ng-class="server.isStopped()?'fa fa-power-off':'pficon-ok'"></span>{{server.name}}
+              <span class="pficon" ng-class="server.isStopped()?'fa fa-power-off':'pficon-ok'"></span>{{server.getHost()}}
```
tristantarrant: I think we should swap them: server name (which is unique across the cluster) in the card title and host name in the card body

vblagoje: Aha, yeah it makes sense. One sec

src/main/webapp/cache-container/configuration-security/security.html
vblagoje: Ok, correcting

tristantarrant:
Thanks Vladimir, pushed with some additional changes.

PR #76, opened 2016-03-23
ISPN-6338 Disabling cache status is lost in the UI

vblagoje:
Master only. PR works and tested only on new 9.0 server and beyond. Please decide when to integrate.


tristantarrant:
Pushed to master, thanks

PR #78, opened 2016-04-06
ISPN-6394 Coalesce server group view and Infinispan/JGroups view

vblagoje:
Preview only. Do not integrate before we discuss the details.

### DIFF
--- a/src/main/webapp/clusters-view/clusters-view.controller.js
+++ b/src/main/webapp/clusters-view/clusters-view.controller.js
@@ -104,6 +104,11 @@ angular.module('managementConsole')
                 $scope.relays[cluster.getName()] = ['N/A'];
               }
               $scope.endpoints.push(cluster.getEndpoints());
+              clusterNodesService.getAvailability(cluster).then(function (result) {
+                cluster.availability = result;
+              }).catch(function () {
+                cluster.availability = 'UNAVAILABLE';
+              });
             });
 
             // Refresh grid events
--- a/src/main/webapp/management-console.js
+++ b/src/main/webapp/management-console.js
@@ -62,7 +62,7 @@ angular.module('managementConsole', [
               templateUrl: 'clusters-view-physical/clusters-view.html',
               controller: 'ClustersViewPhysicalCtrl',
               resolve: {
-                serverGroups: function (modelController, utils, $stateParams) {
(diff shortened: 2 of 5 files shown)


vblagoje:
@tristantarrant I made changes in one commit. But will leave the notes in the code about the details

src/main/webapp/clusters-view/clusters-view.controller.js
```diff
@@ -104,6 +104,11 @@ angular.module('managementConsole')
                 $scope.relays[cluster.getName()] = ['N/A'];
               }
               $scope.endpoints.push(cluster.getEndpoints());
+              clusterNodesService.getAvailability(cluster).then(function (result) {
```
vblagoje: @tristantarrant here we get availability for the cache container

src/main/webapp/management-console.js
```diff
@@ -93,11 +93,10 @@ angular.module('managementConsole', [
                         cluster.nodeCount += host.count;
                       });
 
-                      if (serversInGroup.some(isStopped) && serversInGroup.some(isRunning)) {
-                        cluster.status = 'DEGRADED';
-                      }
-                      else if (serversInGroup.some(isRunning)) {
-                        cluster.status = 'STARTED';
```
vblagoje: @tristantarrant and here for server group. If at least one server is running and all servers have the same coordinator then there is not split. In that case we mark server group 'STARTED'. Otherwise if coords are different we have a split. In that case we mark it as 'DEGRADED'. If no servers in the group are running then it is STOPPED. This needs to be tested - will do

src/main/webapp/management-console.js
```diff
@@ -156,11 +155,10 @@ angular.module('managementConsole', [
                       }
                     });
 
-                    if (serversInGroup.some(isStopped) && serversInGroup.some(isRunning)) {
-                      cluster.status = 'DEGRADED';
-                    }
-                    else if (serversInGroup.some(isRunning)) {
-                      cluster.status = 'STARTED';
+                    if (serversInGroup.some(isRunning)) {
```
vblagoje: Same from the above for the clusterNode view itself

tristantarrant:
Pushed to master and 8.2.x

PR #79, opened 2016-04-11
ISPN-6465 Managent console - launching task with no parameters create…

vblagoje:
I believe this one can go on both branches. Please review and test parameter passing yourself.


tristantarrant:
Pushed to master and 8.2.x

PR #83, opened 2016-04-26
ISPN-6120 Management console - there should be a warning about data loss

vblagoje:
Amended the text in the confirmation dialog.


vblagoje:
Integrated with c883fb3416e37e65f839790982210e43c4f0aefb

PR #84, opened 2016-04-27
ISPN-6540 Management console doesn't display cache containers correct…

vblagoje:
Master only. Will backport but needs extra care as this is a very big commit. DO NOT integrate yet.


vblagoje:
@ryanemerson see if you can find any issues on this one! It is a quite a revamp as we can not just use cache container (cluster) as a unique name but rather a combination of server group name and cache container to give its uniqueness. See https://issues.jboss.org/browse/ISPN-6540 for more details. I rebased the branch with the latest master. If you can't find any issues and automated tests pass (see JIRA) then we can integrate this one!

ryanemerson:
Integrated with 2af86da850c658cf9715c3392f569c910c0d5a1e. Thanks @vblagoje

PR #86, opened 2016-04-28
ISPN-6502: Disable editing cache type.

ryanemerson:
https://issues.jboss.org/browse/ISPN-6502

### DIFF
--- a/src/main/webapp/components/directives/cache/configuration-section.html
+++ b/src/main/webapp/components/directives/cache/configuration-section.html
@@ -10,15 +10,15 @@
           <div class="form-group">
             <label class="col-md-4 control-label">{{resolveFieldName(field)}}</label>
 
-            <div class="col-md-4" ng-init="elementType = resolveFieldType(field)">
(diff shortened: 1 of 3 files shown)


vblagoje:
@ryanemerson we need https://github.com/infinispan/infinispan-management-console/pull/87 integrated before this one. Also, note that when creating a new cache with "none" template we need to allow user to change configuration template type. For all other cases we need to forbid. See if you can adapt this PR to fit that requirement.

ryanemerson:
@vblagoje Disabling changes to template type shouldn't be a problem.  I have noticed that my current solution disables selecting  a cache type on the create cache screen, so I will need to make further changes.

vblagoje:
@ryanemerson ok, let me know when this one is ready for testing and a review!

ryanemerson:
@vblagoje I have pushed a new solution. Note that this is very different to before.

vblagoje:
Thanks @ryanemerson Integrated with 18d59960ceb80c6fcf23c3a510220192874f4645

PR #87, opened 2016-04-28
ISPN-6571 Adapt to configuration template changes

vblagoje:
Master only. @ryanemerson see ISPN-6571 for details. We need this one integrated before others. Please review thoroughly.

### DIFF
--- a/src/main/webapp/components/directives/cache/local-cache.html
+++ b/src/main/webapp/components/directives/cache/local-cache.html
@@ -1,7 +1,7 @@
 <div>
   <uib-tabset vertical="true" type="pills">
     <uib-tab heading="General" active="true">
-      <configuration-section fields="[{name: 'Main', fields: ['type', 'template']}, 
+      <configuration-section fields="[{name: 'Main', fields: ['type', 'template-name']},
                                       {name: 'Monitoring', fields: ['statistics', 'jndi-name']}]" data='data'
                              metadata="resolveDescription('general')" init-defaults="initDefaults" read-only="readOnly"/>
     </uib-tab>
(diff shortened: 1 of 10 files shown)


src/main/webapp/components/directives/cache/local-cache.html
```diff
@@ -34,7 +34,7 @@
       <configuration-section fields="[{name: 'Transactions', fields:['locking', 'mode', 'stop-timeout']}]" data="data.transaction.TRANSACTION"
                              metadata="resolveDescription('tx')" init-defaults="initDefaults" read-only="readOnly"/>
     </uib-tab>
-
```
ryanemerson: Should we avoid unnecessary style changes to improve diffs?

ryanemerson:
@vblagoje Overall it looks good, however you need to update the message in [confirmation-cache-modal.html](https://github.com/infinispan/infinispan-management-console/blob/master/src/main/webapp/edit-cache/confirmation-cache-modal.html).  It currently outputs "Update configuration template false" as it still references ```configurationModel.template```

src/main/webapp/components/directives/cache/local-cache.html
vblagoje: Right @ryanemerson I have to turn off automatic on save formatter in Webstorm.

vblagoje:
Awesome @ryanemerson Man I can't tell you how glad I am to have you on the team and that I am not in this alone!! I updated changes but HTML tab/blank line formatting is still an issue. I added a commit to PR so if you can fix it as you squash commits I would be grateful. I'll adjust Webstorm not to edit these in the future.

ryanemerson:
Integrated with 38d8cc5f892757ea89d921b77ab3b7de35eb41bc. Thanks @vblagoje

PR #88, opened 2016-05-02
ISPN-5968 Cache statistics are not shown for local caches in mgmt con…

vblagoje:
All branches. @ryanemerson please review and integrate. Note that on cache-status.html page we display cluster-wide stats for the cache. However, since local is obviously not cluster-wide we just show its own one node cache stats

### DIFF
--- a/src/main/webapp/cache-status/cache-status.html
+++ b/src/main/webapp/cache-status/cache-status.html
@@ -154,7 +154,7 @@
       </div>
     </div>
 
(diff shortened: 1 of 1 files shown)


vblagoje:
You'll need to create one local test cache @ryanemerson

ryanemerson:
@vblagoje Integrated https://github.com/infinispan/infinispan-management-console/commit/2f54c8372f9839ac6e511766226da8d461281813, thanks!

PR #89, opened 2016-05-03
ISPN-6079 - Searching nodes in management console by IP address doesn…

pzapataf:
Using one filter instead of two is enough and simpler


ryanemerson:
@vblagoje looks good to me.

vblagoje:
Thanks @pzapataf Integrated with c3eaea362fdfa39c5970c945c995486a38e31bb5

PR #96, opened 2016-05-05
Remove status filters from cluster view

vblagoje:
Not a resolution for https://issues.jboss.org/browse/ISPN-5916 but as we are postponing the fix - we should at least remove these filters as they are not working atm


ryanemerson:
Integrated https://github.com/infinispan/infinispan-management-console/commit/013fe25fbc7c39f5ce1a6795f6a5edafb95a2479, thanks @vblagoje

PR #99, opened 2016-05-10
ISPN-6531 Workaround for cache configuration security roles editing

vblagoje:
Master branch. See https://issues.jboss.org/browse/ISPN-6531 for more details

### DIFF
--- a/src/main/webapp/components/api/cache-creation-controller.service.js
+++ b/src/main/webapp/components/api/cache-creation-controller.service.js
@@ -128,6 +128,14 @@ angular.module('managementConsole.api')
           this.createHelper(steps, address.concat('binary-keyed-jdbc-store', 'BINARY_KEYED_JDBC_STORE'), configuration['binary-keyed-jdbc-store']);
           this.createHelper(steps, address.concat('mixed-keyed-jdbc-store', 'MIXED_KEYED_JDBC_STORE'), configuration['mixed-keyed-jdbc-store']);
           this.createHelper(steps, address.concat('backup', 'BACKUP'), configuration.backup);
+
+          //if we are creating new node AUTHORIZATION we have to flag its ancestor SECURITY as new node as well
+          var auth = utils.deepValue(configuration.security, 'SECURITY.authorization.AUTHORIZATION');
+          if (utils.isNotNullOrUndefined(auth) && auth['is-new-node']) {
+            var sec = utils.deepValue(configuration.security, 'SECURITY');
+            sec['is-new-node'] = true;
+          }
+
           this.createHelper(steps, address.concat('security', 'SECURITY'), configuration.security);
           this.createHelper(steps, address.concat('security', 'SECURITY', 'authorization', 'AUTHORIZATION'), configuration.security.SECURITY.authorization);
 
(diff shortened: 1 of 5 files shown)


ryanemerson:
@vblagoje It seems to work for me, with two caveats.

1. There are problems with a null ```obj``` being passed to ```utils.deepValue```. The follow error occurs many times in the console for different properties:
```javascript
angular.js:12416 TypeError: Cannot read property 'SECURITY' of null
    at Object.deepValue (utils.js:142)
    at Cache.isSecured (cache-model.service.js:99)
    at Cache.hasSecurityEnabled (cache-model.service.js:82)
    at Object.fn [as get] (eval at <anonymous> (angular.js:1), <anonymous>:4:254)
    at Scope.$digest (angular.js:15745)
    at angular.js:15955
    at completeOutstandingRequest (angular.js:5490)
    at angular.js:5762
```

2. I think it would be better if the roles loaded from the server were formatted in the same way as they are entered. For example, roles are entered as ```["admin", "reader"]``` but when a config is loaded, they are in the format ```admin, reader```. This could be confusing for the user, as they will assume the provided format is the correct one (even if the tooltip says otherwise :) )

src/main/webapp/components/api/cache-creation-controller.service.js
```diff
@@ -128,6 +128,14 @@ angular.module('managementConsole.api')
...
           this.createHelper(steps, address.concat('mixed-keyed-jdbc-store', 'MIXED_KEYED_JDBC_STORE'), configuration['mixed-keyed-jdbc-store']);
           this.createHelper(steps, address.concat('backup', 'BACKUP'), configuration.backup);
+
+          //if we are creating new node AUTHORIZATION we have to flag its ancestor SECURITY as new node as well
+          var auth = utils.deepValue(configuration.security, 'SECURITY.authorization.AUTHORIZATION');
+          if (utils.isNotNullOrUndefined(auth) && auth['is-new-node']) {
+            var sec = utils.deepValue(configuration.security, 'SECURITY');
+            sec['is-new-node'] = true;
+          }
+
```
ryanemerson: Minor nitpick, this code is identical to the code in updateConfigurationTemplate. Is it worth creating a function that can be called? Or do you expect this code to be removed when we revisit security roles?

src/main/webapp/components/directives/cache/configurationViewer.js
```diff
@@ -56,8 +73,22 @@
...
+          };
+
           scope.resolveDescription = function (elementPath) {
-            return utils.resolveDescription(scope.metadata, scope.resourceDescriptionMap, elementPath, scope.cacheType);
+            var overriden = scope.resolveOverridenDescription(elementPath);
+            if (overriden) {
+              return overriden;
+            } else {
+              var desc = utils.resolveDescription(scope.metadata, scope.resourceDescriptionMap, elementPath, scope.cacheType);
+              return desc;
```
ryanemerson: No need for variable ```desc``` here.

ryanemerson:
Integrated 7daeceedc8689efcd2f591fed2347249d27174a7, thanks @vblagoje

PR #100, opened 2016-05-11
ISPN-6533: not possible to change JDBC store configuration in console.

ryanemerson:
https://issues.jboss.org/browse/ISPN-6533

### DIFF
--- a/src/main/webapp/components/api/cache-creation-controller.service.js
+++ b/src/main/webapp/components/api/cache-creation-controller.service.js
@@ -126,11 +126,11 @@ angular.module('managementConsole.api')
           this.createHelper(steps, address.concat('store', 'STORE'), configuration.store);
           this.createHelper(steps, address.concat('file-store', 'FILE_STORE'), configuration['file-store']);
           this.createHelper(steps, address.concat('leveldb-store', 'LEVELDB_STORE'), configuration['leveldb-store']);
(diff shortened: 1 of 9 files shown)


src/main/webapp/components/api/cache-creation-controller.service.js
```diff
@@ -198,13 +209,34 @@ angular.module('managementConsole.api')
...
 
+      CacheCreationControllerClient.prototype.updateJDBCStore = function(steps, address, configuration) {
+        var jdbcType = configuration['jdbc-type'];
+        if (utils.isNullOrUndefined(jdbcType) || jdbcType.length < 1 || utils.isNullOrUndefined(configuration[jdbcType])) {
+          return;
+        }
+
+        // Add step to create/update JDBC store
+        var objectKey = jdbcType.toUpperCase().replace(/-/g, '_');
+        debugger;
```
vblagoje: @ryanemerson what's debugger for?

vblagoje:
@ryanemerson after user enters table values the fields remain permanently dirty upon subsequent loads (their text field background is yellow instead of white). Everything else seems fine. Please see if this PR works well with https://github.com/infinispan/infinispan-management-console/pull/101 and if so integrate both! And please create PRs for downstream

src/main/webapp/components/api/cache-creation-controller.service.js
ryanemerson: @vblagoje good spot, that definitely should not be here. It's a bit of a workaround for webstorm's debugger.

ryanemerson:
Integrated with 7dc4866e05ceb77a4adaf779ca2c0e0bf8f37a3b

PR #105, opened 2016-05-25
ISPN-6700: Restart required dialog show only after editing distribute…

ryanemerson:
…d cache.

https://issues.jboss.org/browse/ISPN-6700


vblagoje:
Resolved with d412624 Thanks @ryanemerson

PR #107, opened 2016-06-06
ISPN-6632: Enable configuration of db-major and db-minor versions in …

ryanemerson:
…Console

Master only.

https://issues.jboss.org/browse/ISPN-6632


ryanemerson:
Closing as I will just incorporate this as part of ISPN-6777.

PR #108, opened 2016-06-06
ISPN-6759 - Fixing IE issues with '#' links

pzapataf:
Links with href='#' were causing Javascript errors in IE. Removing href attribute solves the issue.

https://issues.jboss.org/browse/ISPN-6759


ryanemerson:
Integrated with 71ad54432149931f8220c74e11bddeaec5cbf210

PR #112, opened 2016-06-16
ISPN-6788 Administration console - additional information in forms is…

vblagoje:
Master only. See https://issues.jboss.org/browse/ISPN-6788


ryanemerson:
@vblagoje Whilst this change makes all of the text visible, it also seems to make it much harder to keep the tooltip active. For some reason the change to tooltip-placement causes the tooltip to close almost immediately when it is hovered over in Chrome.

ryanemerson:
Closing as there is a few issues with this in chrome and firefox, we will come back to this later.

PR #115, opened 2016-07-19
ISPN-6779 Improve Security Configuration screen

vblagoje:
Master only. Do not integrate yet. Feedback and review, please @ryanemerson

### DIFF
--- a/src/main/webapp/components/directives/cache/security/security.js
+++ b/src/main/webapp/components/directives/cache/security/security.js
@@ -117,7 +117,7 @@
         };
 
         scope.cleanMetadata = function () {
-          var fields = ['enabled'];
+          var fields = ['enabled', 'roles'];
           fields.forEach(function (field) {
             scope.cleanFieldMetadata(field);
             if (utils.isNotNullOrUndefined(scope.data[field])) {
--- a/src/main/webapp/components/directives/cache/security/security.js
+++ b/src/main/webapp/components/directives/cache/security/security.js
@@ -1,0 +1,100 @@
+(function () {
+  'use strict';
+
+  var module = angular.module('ispn.directives.cache.security', ['ispn.services.utils']);
+
+  module.directive('security', ['utils', '$modal', function (utils, modal) {
+    return {
+      restrict: 'E',
+      scope: {
+        data: '=',
+        cluster: '=',
+        metadata: '=',
+        initDefaults: '=',
+        readOnly: '=',
+        outsideController: '='
+      },
+      replace: true,
+      templateUrl: 'components/directives/cache/security/security.html',
+      link: function (scope) {
+        if (utils.isNotNullOrUndefined(scope.outsideController)) {
+          if (utils.isArray(scope.outsideController)) {
+            var handle = {};
+            scope.outsideController.push(handle);
+            scope.internalController = handle;
+          } else {
+            scope.internalController = scope.outsideController;
+          }
+        } else {
+          scope.internalController = {};
+        }
+        scope.hasSecurityBeenDefinedGlobally = function () {
+          return utils.isNotNullOrUndefined(scope.cluster.security) && utils.isNotNullOrUndefined(scope.cluster.security.SECURITY.authorization.AUTHORIZATION) ;
+        };
+
+        if (utils.isNullOrUndefined(scope.data)) {
+          scope.data = {};
+          utils.deepSet(scope.data, "SECURITY.authorization.AUTHORIZATION", {});
+          scope.data = scope.data.SECURITY.authorization.AUTHORIZATION;
+        }
+
+        scope.containerRoles = [];
+        if (scope.hasSecurityBeenDefinedGlobally()) {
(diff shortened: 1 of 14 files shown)


src/main/webapp/components/directives/cache/security/security.js
```diff
@@ -0,0 +1,100 @@
...
+            scope.data.roles.push(role);
+          }
+        };
+
+        scope.isRoleSelected = function (role) {
+          return scope.data.roles.indexOf(role) > -1;
+        };
+
+        scope.isAnyFieldModified = function () {
+          return true;
```
ryanemerson: Should this always be true?

ryanemerson:
@vblagoje A few thoughts:

1. As we are now using it's own directive, and such features are possible, should we disable the selecting of roles if security is not currently enabled?

2. As we provide an 'undo change' operation throughout the cache config screens, should we not do the same here? My thoughts are that we should ideally allow this for the role selectors, but at the very least it should be there for the 'enable' checkbox.

vblagoje:
@ryanemerson done on both points. Please review again, squash and integrate!

src/main/webapp/components/directives/cache/security/security.js
```diff
@@ -0,0 +1,158 @@
+(function () {
+  'use strict';
+
+  var module = angular.module('ispn.directives.cache.security', ['ispn.services.utils']);
+
+  module.directive('security', ['utils', '$modal', function (utils, modal) {
```
ryanemerson: $modal is redundant

src/main/webapp/components/directives/cache/security/security.js
```diff
@@ -0,0 +1,158 @@
...
+          scope.internalController = {};
+        }
+        scope.hasSecurityBeenDefinedGlobally = function () {
+          return utils.isNotNullOrUndefined(scope.cluster.security) && utils.isNotNullOrUndefined(scope.cluster.security.SECURITY.authorization.AUTHORIZATION) ;
+        };
+
+        if (utils.isNullOrUndefined(scope.data)) {
+          scope.data = {};
+          utils.deepSet(scope.data, "SECURITY.authorization.AUTHORIZATION", {});
+          scope.data = scope.data.SECURITY.authorization.AUTHORIZATION;
```
ryanemerson: This does not work as expected when we pass the data through to the `cache-creation-controller.service` as we have not set the 'is-new-node' field to true on either `security.SECURITY` or `security.SECURITY.authorization.AUTHORIZATION` nodes.  This results in the following error when we attempt to update an existing config with security.SECURITY undefined:

```WFLYCTL0216: Management resource '[ ("profile" => "clustered"), ("subsystem" => "datagrid-infinispan"), ("cache-container" => "clustered"), ("configurations" => "CONFIGURATIONS"), ("distributed-cache-configuration" => "default"), ("security" => "SECURITY") ]' not found```

File: src/main/webapp/components/directives/cache/security/security.js [hunk 4]
```diff
@@ -0,0 +1,158 @@
...
+          scope.metadata[field].uiModified = false;
+          scope.metadata[field].style = null;
+        };
+
+        scope.fieldChangeRequiresRestart = function (field) {
+          return utils.isNotNullOrUndefined(scope.metadata[field]) && scope.metadata[field]['restart-required'] !== 'no-services';
+        };
+
+        scope.requiresRestart = function () {
+          return false;
```
ryanemerson: Should this be false? The resource description states `"restart-required" => "resource-services"` for both the enabled and roles attributes.

src/main/webapp/components/directives/cache/security/security.js
```diff
@@ -0,0 +1,158 @@
...
+        };
+
+        scope.undoFieldChange = function (field) {
+          scope.data[field] = scope.prevData[field];
+          scope.metadata[field].uiModified = false;
+          scope.metadata[field].style = null;
+        };
+
+        scope.fieldChangeRequiresRestart = function (field) {
+          return utils.isNotNullOrUndefined(scope.metadata[field]) && scope.metadata[field]['restart-required'] !== 'no-services';
```
ryanemerson: Currently this is not utilised anywhere, but I think it should be based upon my comments below. The template will also need updating to reflect this.

src/main/webapp/components/directives/cache/security/security.js
```diff
@@ -0,0 +1,158 @@
...
+
+  module.directive('security', ['utils', '$modal', function (utils, modal) {
+    return {
+      restrict: 'E',
+      scope: {
+        data: '=',
+        cluster: '=',
+        metadata: '=',
+        initDefaults: '=',
+        readOnly: '=',
```
ryanemerson: Minor, but we don't actually use `initDefaults` or `readOnly` at present. Maybe we should utilise readOnly for the checkboxes ng-disabled field? Currently I don't think initDefaults is applicable as we only have checkbox inputs in this directive.

ryanemerson:
@vblagoje Integrated ed88e77bd9db08acd5d5791e670f90177d2404c6, thanks.

PR #116, opened 2016-07-22
ISPN-6777 and ISPN-6778: Cache store and loader updated

ryanemerson:
JIRA: https://issues.jboss.org/browse/ISPN-6777

@vblagoje all of the stores are feature complete based upon the design doc. Let me know what you think


ryanemerson:
@vblagoje I have also added ISPN-6778 to this PR in order to reduce merge conflicts. This change is much smaller though and shouldn't require too much testing. Currently ISPN-6778 only has a single cache loader enumerated, so once I have clarified which loaders exactly should be included, I will push an updated list. This shouldn't effect testing etc though, as the loader selector simply returns a class string.

vblagoje:
ok @ryanemerson will take a look today and we'll integrate tomorrow.

src/main/webapp/components/api/cache-creation-controller.service.js
```diff
@@ -366,8 +423,11 @@ angular.module('managementConsole.api')
             var propKey = keys[i];
             var propValue = prop[keys[i]];
             if (utils.isNotNullOrUndefined(propValue)) {
-              if (utils.isObject(propValue) && allowedObjects.indexOf(propKey) > -1) {
-                op[propKey] = propValue;
+              if (utils.isObject(propValue)) {
+                // Only process allowed objects, these should be objects which are dmr attributes not children
+                if (allowedObjects.indexOf(propKey) > -1) {
+                  op[propKey] = propValue;
+                }
```
ryanemerson: @vblagoje  FYI: It's necessary to nest the second check here, otherwise the objects are simply added further down as part of the if statement. This was a problem that I introduced several commits ago which had gone unnoticed, but resulted in lots of crap being sent to the server unnecessarily.

ryanemerson:
@vblagoje Locally I have been working on ISPN-6889 (create generic form-groups directive), which reduces some of the duplicated directive code that is currently found in this PR. It also significantly reduces the amount of html required in the templates. Have a look at [this](https://github.com/infinispan/infinispan-management-console/compare/master...ryanemerson:ISPN-6889) branch and let me know whether you think it's worth integrating as part of this PR.

vblagoje:
Integrated with 771581588a3888f0e5de03a96526505983297f62 and 38c90e9243c0cb41f70afd6d347c5c39707bdf3f . Fantastic work @ryanemerson

PR #117, opened 2016-07-25
ISPN-6888: Enhance field undo/restart required behaviour.

ryanemerson:
- A field is now only classified as modified if its latest is value is not equal to the value loaded during initialisation.

JIRA: https://issues.jboss.org/browse/ISPN-6888


vblagoje:
Integrated with 303a9ce Thanks @ryanemerson

PR #119, opened 2016-07-29
ISPN-6889: Create directive for generic field form groups.

ryanemerson:
-- Directives also refactored to make use of generic code that is now available in utils.js

https://issues.jboss.org/browse/ISPN-6891


vblagoje:
Integrated with 525258f451f6ef08c51dd44b6ad6254a57af1667 Great work and efforrt @ryanemerson

PR #123, opened 2016-08-08
ISPN-6934: Refactor the console build system

ryanemerson:
https://issues.jboss.org/browse/ISPN-6934


tristantarrant:
Building this with mvn clean install creates an empty jar

ryanemerson:
@tristantarrant There are a few problems with the build in general, so I am going to work on them and re-push

ryanemerson:
@tristantarrant The build is fixed and we now have minification support, so the resulting jar is approximately half the size as before. I have also fixed various problems with images not being loaded and file translations should work as expected  now.

ryanemerson:
@vblagoje This should only be merged if we don't go down the Typescript route, as the build process changes if we do.

ryanemerson:
Closing as the TypeScript refactor makes this redundant.

PR #125, opened 2016-09-05
New server instance page

vblagoje:
Please integrate into TypeScriptFresh @ryanemerson . Yes, there is an extra function for getCoordinator but let's keep it for now and we can remove it later. Feedback is appreciated!

### DIFF
--- a/src/module/server-instance/MemoryChart.ts
+++ b/src/module/server-instance/MemoryChart.ts
@@ -1,0 +1,34 @@
+/// <reference path="../../../typings/globals/d3/index.d.ts" />
+/// <reference path="../../../typings/globals/c3/index.d.ts" />
+
+import * as c3 from "c3";
+export class InstanceMemoryChart {
+    chart: c3.ChartAPI;
+
+    constructor(bindTo: string, used: number, max: number) {
+        this.chart = c3.generate({
+            bindto: bindTo,
+            size: {
+                width: 220,
+                height: 220
+            },
+            data: {
+                columns: [["Used", used], ["Free", max]],
+                type: "donut"
+            },
+            donut: {
+                title: "Memory"
+            }
+
+        });
+    }
+
+    redraw(): void {
+        this.chart.flush();
+    }
+
+    destroy(): void {
+        this.chart.destroy();
+    }
+}
+
--- a/src/module/server-instance/ServerInstance.ts
+++ b/src/module/server-instance/ServerInstance.ts
@@ -1,0 +1,45 @@
+import {App} from "../../ManagementConsole";
+import {ServerInstanceCtrl} from "./ServerInstanceCtrl";
+import {ServerAddress} from "../../services/server/ServerAddress";
+import {IServerAddress} from "../../services/server/IServerAddress";
+import {IStateParamsService} from "angular-ui-router";
+import {ServerGroupService} from "../../services/server-group/ServerGroupService";
+import {JGroupsService} from "../../services/jgroups/JGroupsService";
+import {ServerService} from "../../services/server/ServerService";
+
+const module: ng.IModule = App.module("managementConsole.server-instance", []);
+
+module.controller("ServerInstanceCtrl", ServerInstanceCtrl);
+
+// @ngInject
+module.config(($stateProvider: ng.ui.IStateProvider) => {
+  $stateProvider.state("server-instance", {
+    parent: "root",
+    url: "server-groups/:serverGroupName/:host/:server",
+    templateUrl: "module/server-instance/view/server-instance.html",
+    controller: ServerInstanceCtrl,
+    controllerAs: "ctrl",
+    resolve: {
+      /* tslint:disable:no-string-literal */
+      coord: ["$q", "$stateParams", "serverGroupService", "jGroupsService", "serverService",
+        ($q: ng.IQService, $stateParams: IStateParamsService, serverGroupService: ServerGroupService,
+         jGroupsService: JGroupsService, serverService: ServerService) => {
+          let deferred: ng.IDeferred<IServerAddress> = $q.defer<IServerAddress>();
+          let server: IServerAddress = new ServerAddress($stateParams["host"], $stateParams["server"]);
+          serverService.getServerStatus(server).then(status => {
+            if (status === "STOPPED") {
+              deferred.resolve(new ServerAddress("", ""));
+            } else {
+              return serverGroupService.getServerGroupMapWithMembers($stateParams["serverGroupName"]).then(resp => {
+                deferred.resolve(jGroupsService.getCoordinatorByServer(server, resp.profile));
+              });
+            }
+          });
+          return deferred.promise;
+        }],
+      serverInstance: ["$stateParams", "serverService", ($stateParams: IStateParamsService, serverService: ServerService) => {
+        return serverService.getServer(new ServerAddress($stateParams["host"], $stateParams["server"]));
+      }]
+    }
+  });
+});
--- a/src/module/server-instance/ServerInstanceCtrl.ts
+++ b/src/module/server-instance/ServerInstanceCtrl.ts
@@ -1,0 +1,120 @@
+import {deepGet} from "../../common/utils/Utils";
+import {openConfirmationModal} from "../../common/dialogs/Modals";
+import {ServerService} from "../../services/server/ServerService";
+import {IServerAddress} from "../../services/server/IServerAddress";
+import {IServer} from "../../services/server/IServer";
+import {IIntervalService} from "angular";
+import {InstanceMemoryChart} from "./MemoryChart";
+import IModalServiceInstance = angular.ui.bootstrap.IModalServiceInstance;
+import IModalService = angular.ui.bootstrap.IModalService;
+import {IStateService} from "angular-ui-router";
+
+export class ServerInstanceCtrl {
+  static $inject: string[] = ["$state", "$interval", "$uibModal", "serverService", "coord", "serverInstance"];
+
+  private threadCount: number;
+  private threadPeakCount: number;
+  private threadDaemonCount: number;
+
+  private directBufferPoolCount: number;
+  private directBufferPoolMemoryUsed: number;
+
+  private mappedBufferPoolCount: number;
+  private mappedBufferPoolMemoryUsed: number;
+
+  private nodeStats: any;
+  private chart: InstanceMemoryChart;
+
+  constructor(private $state: IStateService, private $interval: IIntervalService, private $uibModal: IModalService,
+              private serverService: ServerService, private coord: IServerAddress, private serverInstance: IServer) {
+    this.$interval(() => (this.refreshStats()), 500, 1);
+  }
+
+  refreshStats(): void {
+    let server: IServerAddress = this.serverInstance.address;
+    this.fetchThreadAndMemoryStats(server);
+    this.fetchAggregateNodeStats(server);
+  }
+
+  fetchThreadAndMemoryStats(address: IServerAddress): void {
+    this.serverService.getServerStats(address).then(response => {
+      // memory
+      let memory: any = response.memory["heap-memory-usage"];
+      let used: number = (memory.used / 1024) / 1024;
+      let max: number = (memory.max / 1024) / 1024;
+
+      // threading
+      let threading: any = response.threading;
+      this.threadCount = threading["thread-count"];
+      this.threadPeakCount = threading["peak-thread-count"];
+      this.threadDaemonCount = threading["daemon-thread-count"];
+
+      let directBufferPool: any = deepGet(response, "buffer-pool.name.direct");
+      let mappedBufferPool: any = deepGet(response, "buffer-pool.name.mapped");
+
+      this.directBufferPoolCount = directBufferPool.count;
+      this.directBufferPoolMemoryUsed = directBufferPool["memory-used"];
+
+      this.mappedBufferPoolCount = mappedBufferPool.count;
+      this.mappedBufferPoolMemoryUsed = mappedBufferPool["memory-used"];
+
+      if (this.chart) {
+        this.chart.destroy();
+      }
+      this.chart = new InstanceMemoryChart("#chart", used, max);
+    });
+  }
+
+  fetchAggregateNodeStats(address: IServerAddress): void {
+
+    this.serverService.getAggregateNodeStats(address).then(response => {
+      // TODO here we need to loop through all cache containers and add all stats up
+      // but for now just use the first container found
+      let containersRoot: any = response["cache-container"];
+      for (var prop in containersRoot) {
+        this.nodeStats = containersRoot[prop];
+        break;
+      }
+    });
+  }
+
+  isCoordinator(): boolean {
+    return this.serverInstance.address.host === this.coord.host && this.serverInstance.address.name === this.coord.name;
+  };
+
+  createStartModal(): void {
+    let modal: IModalServiceInstance = openConfirmationModal(this.$uibModal, "Start server " + this.serverInstance.address.name + "?");
+    modal.result.then(() => {
+      let promise: ng.IPromise<string> = this.serverService.startServer(this.serverInstance.address);
+      promise.then(() =>
+        this.refresh()
+      );
+    });
+  }
+
+  createStopModal(): void {
+    let modal: IModalServiceInstance = openConfirmationModal(this.$uibModal, "Stop server " + this.serverInstance.address.name + "?");
+    modal.result.then(() => {
+      let promise: ng.IPromise<string> = this.serverService.stopServer(this.serverInstance.address);
+      promise.then(() =>
+        this.refresh()
+      );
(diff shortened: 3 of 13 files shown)


ryanemerson:
@vblagoje You need to include the updated typings.json otherwise the d3 and c3 typings are not installed. Also if you have changed the dependencies, now that we don't use c3-angular, an updated package.json/jspm.conf should be included.

src/module/server-instance/MemoryChart.ts
```diff
@@ -0,0 +1,33 @@
...
+            size: {
+                width: 220,
+                height: 220
+            },
+            data: {
+                columns: [['Used', used], ['Free', max]],
+                type: 'donut'
+            },
+            donut: {
+                title:'Memory'
```
ryanemerson: Minor nitpick, but lets try to use double quotes in Typescript files, as per the rules of TSlint.  You can set webstorm to make these checks for you. Also, we should be able to call gulp check on the cmd line, however this isn't currently working, so I will look into that

src/module/server-instance/ServerInstance.ts
```diff
@@ -0,0 +1,37 @@
...
+    controllerAs: "ctrl",
+    params: {
+      serverGroupName: null,
+      host: null,
+      server: null
+    },
+    resolve: {
+      coord: ["$stateParams", "serverGroupService", ($stateParams, serverGroupService) => {
+        var server = new ServerAddress($stateParams.host, $stateParams.server);
+        return serverGroupService.getCoordinator(server, "clustered");
```
ryanemerson: This is assuming that the user is using a profile called "clustered", resulting in the page not displaying if they are using a different profile name. We can get the profile name from the serverGroup. So I think it is best to resolve the serverGroup and then resolve the coordinator by passing "serverGroup" as a dependency to the "coord" resolve function.

src/module/server-instance/ServerInstanceCtrl.ts
```diff
@@ -0,0 +1,85 @@
...
+    directBufferPoolCount: number;
+    directBufferPoolMemoryUsed: number;
+
+    mappedBufferPoolCount: number;
+    mappedBufferPoolMemoryUsed: number;
+
+    nodeStats: any;
+    chart:InstanceMemoryChart;
+
+    constructor(private $interval: IIntervalService, private $stateParams: IStateParamsService,
```
ryanemerson: There's no need to pass $stateParams into the controller as you are only using it to get information that has already been passed to the controller.

src/module/server-instance/ServerInstanceCtrl.ts
```diff
@@ -0,0 +1,85 @@
...
+                private serverService: ServerService, public coord: IServerAddress,
+                private serverGroup:IServerGroup, private serverInstance:IServer) {
+        this.$interval(()=>(this.refreshStats()),500,1);
+    }
+
+    refreshStats(): void {
+        let server = new ServerAddress(this.$stateParams['host'], this.$stateParams['server']);
```
ryanemerson: Here we can just call this.serverInstance.address

src/module/server-instance/view/server-instance.html
```diff
@@ -0,0 +1,163 @@
+<div class="container-fluid">
+  <div class="row header-row">
+    <div class="col-sm-12 col-md-12">
+      <ol class="breadcrumb">
+        <li>
+          <a href="#/clusters-view-physical" translate>CLUSTERS</a>
+        </li>
+        <li><a ui-sref="server-group({serverGroup: ctrl.$stateParams.serverGroupName})" ui-sref-opts="{reload:true}">{{ctrl.$stateParams.serverGroupName}}</a></li>
+        <li class="active"><strong>{{ctrl.$stateParams.server}}</strong></li>
```
ryanemerson: Replace with ctrl.serverGroup.name and ctrl.serverInstance.address.name (or create explicit address/name field, might make the template cleaner)

src/module/server-instance/ServerInstanceCtrl.ts
```diff
@@ -0,0 +1,85 @@
...
+        this.fetchThreadAndMemoryStats(server);
+        this.fetchAggregateNodeStats(server);
+    }
+
+    fetchThreadAndMemoryStats(address: IServerAddress): void {
+        this.serverService.getServerStats(address).then(response => {
```
vblagoje: How to TSLint response?

src/services/server/ServerService.ts
```diff
@@ -39,6 +43,26 @@ export class ServerService {
...
   }
 
+  getServer(server: IServerAddress): ng.IPromise<IServer> {
+    let deferred: ng.IDeferred<IServer> = this.$q.defer<IServer>();
+    this.dmrService.readAttribute({
+      address: this.generateAddress(server),
+      name: "server-state"
+    }).then(serverState =>{
+      if (serverState === "STOPPED") {
+        deferred.resolve(new Server(server, serverState, ''));
```
ryanemerson: Should we not just reject the promise at this point with an error message? It does not make sense to me to create a server object when there isn't one available. In the controller we can then just handle a rejected promise and maybe even display the rejected message?

vblagoje:
Thanks @ryanemerson Your feedback was so valuable. I even simplified things further! Cheers

src/services/server-group/ServerGroupService.ts
```diff
@@ -163,6 +163,34 @@ export class ServerGroupService {
...
+  getChannelName(profile:string): ng.IPromise<string[]> {
+
+    let request: IDmrRequest = <IDmrRequest> {
+      address :['profile', profile, 'subsystem', 'datagrid-jgroups'],
+      'child-type': 'channel'
+    };
+    return this.dmrService.readChildName(request);
+  }
+
+  getCoordinator(server:IServerAddress, profile:string): ng.IPromise<IServerAddress> {
```
ryanemerson: We should really replace this with a call to JGroupsServie.getCoordinatorByServer as they are the same and then delete getChannelName(). It doesn't make sense to duplicate code

src/module/server-instance/ServerInstanceCtrl.ts
```diff
@@ -0,0 +1,81 @@
...
+                this.nodeStats = containersRoot[prop];
+                break;
+            }
+        });
+
+    }
+
+    isCoordinator(server: IServer): boolean {
+        return server.address.host === this.coord.host && server.address.name === this.coord.name;
+    };
```
ryanemerson: No need to pass server instance here as we already have it stored in the controllers scope as this.serverInstance.

src/module/server-instance/ServerInstanceCtrl.ts
```diff
@@ -0,0 +1,81 @@
...
+            if (this.chart) {
+                this.chart.destroy();
+            }
+            this.chart = new InstanceMemoryChart("#chart", used, max);
+        });
+    }
+
+    fetchAggregateNodeStats(address: IServerAddress): void {
+
+        this.serverService.getAggregateNodeStats(address).then(response => {
```
ryanemerson: What happens here if this.serverInstance is stopped? this is possible as ServerService.getServer() simply returns a new Server instance when it a server has been stopped [1] and we will get errors

[1] https://github.com/vblagoje/infinispan-management-console/blob/1e887087d963c11032fd8d3c29c01d78bcfdb817/src/services/server/ServerService.ts#L52

src/services/server/ServerService.ts
```diff
@@ -39,6 +41,27 @@ export class ServerService {
     });
   }
 
+  getServer(server: IServerAddress): ng.IPromise<IServer> {
+    let deferred: ng.IDeferred<IServer> = this.$q.defer<IServer>();
+    this.dmrService.readResource({
+      address: this.generateAddress(server),
+      "include-runtime": true
+    }).then(server => {
```
ryanemerson: What happens if the server is not active here? We should check the server status before proceeding and take required action if it stopped.

src/module/server-instance/ServerInstance.ts
```diff
@@ -0,0 +1,35 @@
...
+        parent: "root",
+        url: "server-groups/:serverGroupName/:host/:server",
+        templateUrl: "module/server-instance/view/server-instance.html",
+        controller: ServerInstanceCtrl,
+        controllerAs: "ctrl",
+        params: {
+            serverGroupName: null,
+            host: null,
+            server: null
+        },
```
ryanemerson: nitpick, but we can get rid of lines 18-22 as they are implicitly declared in the url field

src/module/server-instance/view/server-instance.html
```diff
@@ -0,0 +1,163 @@
...
+        <button class="btn btn-default" type="button" id="refreshButton" ng-click="ctrl.refreshStats()">
+          <span class="pficon pficon-refresh"></span> <span translate>REFRESH</span>
+        </button>
+        <button class="btn btn-default dropdown-toggle" type="button" id="dropdownMenu"
+                data-toggle="dropdown" uib-dropdown-toggle><span translate>ACTIONS</span><span class="caret"></span>
+        </button>
+        <ul class="dropdown-menu" role="menu" aria-labelledby="dropdownMenu">
+          <li role="presentation" ng-class="ctrl.serverInstance.isRunning()?'disabled':''"><a id="dropdownMenuStart" role="menuitem" ng-click="openModal('start')" translate>START</a></li>
+          <li role="presentation" ng-class="!ctrl.serverInstance.isRunning()?'disabled':''"><a id="dropdownMenuStop" role="menuitem" ng-click="openModal('stop')" translate>STOP</a></li>
+          <li role="presentation" ng-class="ctrl.serverInstance.isRunning()?'disabled':''"><a id="dropdownMenuStart" role="menuitem" ng-click="openModal('remove')" translate>REMOVE</a></li>
```
ryanemerson: The stop/start/remove methods to be implemented

src/services/server/ServerService.ts
vblagoje: Well, I am not sure @ryanemerson. Server can be STOPPED and we might still want to read information about it. Why not? I would leave this as is and change it to rejecting promise if we really need to.

ryanemerson: @vblagoje Ok, that makes sense. However the only operations that can be executed on a stopped server, AFAIK, are get launch type and get status, but I guess this could change in the future.

src/module/server-instance/ServerInstanceCtrl.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    isCoordinator(): boolean {
+        return this.serverInstance.address.host === this.coord.host && this.serverInstance.address.name === this.coord.name;
+    };
+
+    createStartModal(): void {
+        let modal: IModalServiceInstance = openConfirmationModal(this.$uibModal, "Start server " + this.serverInstance.address.name + "?");
+        modal.result.then(() => {
+            let promise: ng.IPromise<string> = this.serverService.startServer(this.serverInstance.address);
+            promise.then(() => {}, error => {});
```
ryanemerson: In all of the models, the resolved and reject callbacks are empty. What should we be doing here? Reload the page or remove the empty callbacks?

ryanemerson:
@vblagoje Overall it looks good, we just need to adapt the behaviour of getCoordinatorByServer and the modal issue.

src/module/server-instance/ServerInstanceCtrl.ts
ryanemerson: @vblagoje If a server instance is stopped, then this throws error on page load. For example, if we have a running ServerInstance, which we then stop in the console, when the page is reloaded we get DMR errors from the server stating "No resource definition is registered for address". In the service method, we need to check the status of the server before attempting to retrieve the stats.

src/services/server/ServerService.ts
```diff
@@ -66,6 +105,34 @@ export class ServerService {
...
+      address: [].concat("host", server.host, "server", server.name, "core-service", "platform-mbean"),
+      recursive: true,
+      "child-type": "type",
+      "include-runtime": true,
+      "recursive-depth": 2
+    };
+    this.dmrService.readChildResources(request).then(
+        (response) => deferred.resolve(response),
+        () => deferred.reject());
+    return deferred.promise;
```
ryanemerson: First call this.getServerStatus() and if not stopped try to retrieve the stats. Simply rejecting the deferred promise if the status is stopped should be sufficient IMO.

src/services/server/ServerService.ts
```diff
@@ -66,6 +105,34 @@ export class ServerService {
...
+      address: [].concat("host", server.host, "server", server.name, "subsystem", "datagrid-infinispan", "cache-container"),
+      "include-runtime": true,
+      recursive: true
+    };
+    this.dmrService.readResource(request).then(
+        (response) => deferred.resolve(response),
+        () => deferred.reject());
+    return deferred.promise;
+  }
+
```
ryanemerson: Same as with getServerStats

src/services/server/ServerService.ts
```diff
@@ -39,6 +57,27 @@ export class ServerService {
     });
   }
 
+  getServer(serverAddress: IServerAddress): ng.IPromise<IServer> {
+    let deferred: ng.IDeferred<IServer> = this.$q.defer<IServer>();
+    this.dmrService.readResource({
+      address: this.generateAddress(serverAddress),
+      "include-runtime": true
+    }).then(server => {
+      let serverState:string = server["server-state"].toUpperCase();
```
ryanemerson: lets be consistent and use the getServerStatus method here. It will reduce this part from 5 lines of code to 1.

src/module/server-instance/ServerInstanceCtrl.ts
```diff
@@ -0,0 +1,120 @@
...
+    };
+
+    createStartModal(): void {
+        let modal: IModalServiceInstance = openConfirmationModal(this.$uibModal, "Start server " + this.serverInstance.address.name + "?");
+        modal.result.then(() => {
+            let promise: ng.IPromise<string> = this.serverService.startServer(this.serverInstance.address);
+            promise.then(() => {
+                this.refresh();
+            });
+        });
```
ryanemerson: FYI, we can reduce this further to `promise.then() => this.refresh);`, or even do away with the promise variable. But whatever you prefer.

ryanemerson:
@vblagoje We definitely need to remedy the service calls that throw errors when an instance has stopped, but other than that mainly house keeping tasks. Also, can we remove "server-instance.less" as it is empty?

src/module/server-instance/ServerInstance.ts
```diff
@@ -0,0 +1,38 @@
...
+// @ngInject
+module.config(($stateProvider: ng.ui.IStateProvider) => {
+    $stateProvider.state("server-instance", {
+        parent: "root",
+        url: "server-groups/:serverGroupName/:host/:server",
+        templateUrl: "module/server-instance/view/server-instance.html",
+        controller: ServerInstanceCtrl,
+        controllerAs: "ctrl",
+        resolve: {
+            coord: ["$q", "$stateParams", "serverGroupService", "jGroupsService", "serverService", ($q, $stateParams, serverGroupService, jGroupsService, serverService) => {
```
ryanemerson: We can prevent the compiler errors on $q by defining it as `$q: ng.IQService` in the parameters

src/services/server/ServerService.ts
vblagoje: Okay, I can do that @ryanemerson

src/services/server/ServerService.ts
vblagoje: I know but this call is a bit specific because I can read everything I need to return from the given server object. So we would make two calls here instead of one.

src/module/server-instance/ServerInstanceCtrl.ts
vblagoje: How, I am curious! I cannot make it compile?!

src/services/server/ServerService.ts
ryanemerson: Sorry I missed that you were getting the server group and profile. This is definitely the better way

src/module/server-instance/ServerInstanceCtrl.ts
ryanemerson: Slight typo in my original response, I missed some brackets, but `promist.then( () => this.refresh()); should work.

src/module/server-instance/ServerInstanceCtrl.ts
vblagoje: Done

src/module/server-instance/ServerInstanceCtrl.ts
vblagoje: Aha , no brackets? Done

vblagoje:
Should be ready to go now @ryanemerson . Have a look

ryanemerson:
@vblagoje Integrate https://github.com/infinispan/infinispan-management-console/commit/d85e311a195cf141ebc3ce161f1106b61d6660c1, thanks!

PR #126, opened 2016-09-12
Cache instance page

vblagoje:
Preview for @ryanemerson . Please review :-)

### DIFF
--- a/src/services/cache/CacheService.ts
+++ b/src/services/cache/CacheService.ts
@@ -6,16 +6,20 @@ import {IDmrRequest} from "../dmr/IDmrRequest";
 import {Cache} from "./Cache";
 import {ICacheConfiguration} from "./ICacheConfiguration";
 import {isNullOrUndefined, convertCacheAttributeIntoFieldName} from "../../common/utils/Utils";
+import {IServerAddress} from "../server/IServerAddress";
+import {ICacheContainer} from "../container/ICacheContainer";
+import {JGroupsService} from "../jgroups/JGroupsService";
 
 const module: ng.IModule = App.module("managementConsole.services.cache", []);
 
 export const CACHE_TYPES: string[] = ["distributed-cache", "replicated-cache", "local-cache", "invalidation-cache"];
 
 export class CacheService {
-  static $inject: string[] = ["$q", "dmrService", "launchType"];
+  static $inject: string[] = ["$q", "dmrService", "jGroupsService", "launchType"];
 
   constructor(private $q: ng.IQService,
               private dmrService: DmrService,
+              private jgroupsService: JGroupsService,
               private launchType: LaunchTypeService) {
   }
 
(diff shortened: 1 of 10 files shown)


vblagoje:
@ryanemerson I decided *not* to use sub views as the menu action buttons are completely different on these two pages. I implemented it the way we had it before. Do not see any major downsides to this approach. See if you can find some. Thank you

src/services/cache/CacheService.ts
```diff
@@ -79,6 +92,123 @@ export class CacheService {
     return this.dmrService.readResource(request);
   }
 
+  getCacheStats(container: ICacheContainer, cache: ICache): ng.IPromise<any> {
+    let firstServer: IServerAddress = container.serverGroup.members[0];
+    let address: string [] = [].concat("host", firstServer.host, "server", firstServer.name, "subsystem",
+      "datagrid-infinispan", "cache-container", container.name, cache.type, cache.name);
```
ryanemerson: Minor, but let's make use of the generateAddress function here, like you have in `getCacheUsingType(..)` as this function will eventually return a different address when running in Standalone mode. The same applies to the other request addresses.

src/services/cache/CacheService.ts
```diff
@@ -79,6 +92,123 @@ export class CacheService {
...
+      return this.dmrService.executePost({
+        operation: "cache-rebalance",
+        address: [].concat("host", coord.host, "server", coord.name,
+          "subsystem", "datagrid-infinispan", "cache-container", container.name, cache.type, cache.name),
+        value: rebalance.toString()
+      });
+    });
+  }
+
+  private getAllCachesInContainerShallow(container: string, profile?: string): ng.IPromise<ICache[]> {
```
ryanemerson: This method doesn't appear to be used anywhere

src/services/cache/CacheService.ts
```diff
@@ -70,6 +74,15 @@ export class CacheService {
     return deferred.promise;
   }
 
+  getCacheUsingType(name: string, type: string, container: string, profile?: string): ng.IPromise<ICache> {
```
ryanemerson: It seems that `getCache(..)` is now redundant (I couldn't find any usages and I think it's broken anyway). So we should remove `getCache(..)` all together and maybe rename this to `getCache()`

ryanemerson:
@vblagoje Overall this looks good, but there are errors thrown when loading the nodes page that need fixing. 

There is also errors thrown from the server when executing the "reset statistics" operations. 

Another minor thing. It would be nice if the navigation links at the top of the page also included the caches page. E.g. `Cache containers >> cache-container >> caches >> default (distributed-cache)

src/services/cache/CacheService.ts
vblagoje: +1 makes sense

vblagoje: But @ryanemerson but generateAddress is using profile path and here we need server path!

src/services/cache/CacheService.ts
vblagoje: +1 Correct @ryanemerson

src/services/cache/CacheService.ts
vblagoje: Corrected @ryanemerson

src/services/cache/CacheService.ts
ryanemerson: O yeah, I missed that, maybe we should create a generic function to create the server path then? Also, does it matter which node we get the stats from? AFAIK, it shouldn't but wasn't sure if going via the coordinator would make any difference.

vblagoje:
@ryanemerson I updated PR. I did not change navigations links (except enabling cache containers link) as I think we can deal with this separately later. Have another look when you get a chance?

ryanemerson:
@vblagoje Integrated 3b073182e79456a968363944eb0306a63cf07982, thanks :)

PR #128, opened 2016-09-29
Avoid double click dropdown invocation

vblagoje:
TypeScriptFresh only. Avoids double click dropdown invocation @ryanemerson


vblagoje:
Wrong branch @ryanemerson Will open another PR

PR #136, opened 2016-10-27
Introduce memory chart directive

vblagoje:
@ryanemerson we will use memory chart directive in cache, server instance, and new memory usage tab. _Do not_ integrate as is. See how it works, check it out and when we are ready we will remove the topmost commit that uses memory chart directive to draw memory chart on server instance page.

### DIFF
--- a/src/ManagementConsole.ts
+++ b/src/ManagementConsole.ts
@@ -33,6 +33,7 @@ import IAngularEvent = angular.IAngularEvent;
 import {FileModelDirective} from "./components/directives/FileModelDirective";
 import {VertilizeContainerDirective} from "./components/directives/VertilizeContainerDirective";
 import {VertilizeDirective} from "./components/directives/VertilizeDirective";
+import {MemoryDirective} from "./components/directives/memory/MemoryDirective";
 
 const App: ng.IAngularStatic = angular;
 
@@ -54,6 +55,7 @@ module.config(($translateProvider: ITranslateProvider) => {
 module.directive("fileModel", FileModelDirective.factory());
 module.directive("vertilizeContainer", VertilizeContainerDirective.factory());
 module.directive("vertilize", VertilizeDirective.factory());
+module.directive("memoryChart", MemoryDirective.factory());
 
 // @ngInject
 module.config(($urlRouterProvider: ng.ui.IUrlRouterProvider) => {
--- a/src/components/directives/memory/MemoryChart.ts
+++ b/src/components/directives/memory/MemoryChart.ts
@@ -1,0 +1,43 @@
+/// <reference path="../../../../typings/globals/d3/index.d.ts" />
+/// <reference path="../../../../typings/globals/c3/index.d.ts" />
+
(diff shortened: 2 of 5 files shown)


ryanemerson:
@vblagoje This looks a lot better, great job!

I think that it would be better if we tried to use components instead of directives for new code though, as this will help the transition to angular 2.0 when required and we also seem to have a DI problem with directives and minification.  The obvious problem I can see with changing this from a directive to a component is the use of $watch, however this can be overcome by doing something similar to [this](https://stackoverflow.com/questions/35534479/angularjs-1-5-component-does-not-support-watchers-what-is-the-work-around).

vblagoje:
@ryanemerson seems like ng-change can only be used on HTML input components. See https://docs.angularjs.org/api/ng/directive/ngChange for more details.

ryanemerson:
@vblagoje What are the situations where `data` changes? Is it only on page refresh?

vblagoje:
@ryanemerson right, and on refresh initiated by the user! Maybe we can use onChanges and doCheck?

ryanemerson:
@vblagoje The component object is recreated on a page refresh, so we shouldn't need to do anything

vblagoje:
@ryanemerson ok awesome. Thanks for these insights.

ryanemerson:
@vblagoje This is looking really good :+1: 

A couple of issues I've noticed:

- The directive does not work with minification as we are getting the same DI issue that we get with the other directives.

- The pie chart is not generated on the cache page, only the JVM stats are present.

vblagoje:
Right right @ryanemerson - thank you. I split commits into two branches: https://github.com/vblagoje/infinispan-management-console/tree/memory_all and https://github.com/vblagoje/infinispan-management-console/tree/t_memory_directive I fixed the pie chart on cache page and now we just have to figure out how to convert directives into components. My thoughts are: let's be feature complete and then we can convert directives into components. All of them. I'll work on that next.

ryanemerson:
@vblagoje Ok sounds good. Minor, but I also noticed that we have a few more errors on `gulp check` due to unused variables etc. Let me know once you've pushed the next version for review

vblagoje:
@ryanemerson I updated memory_all branch - cleaned up gulp check errors. Let's just wait for Monday to see how and when DMR ops will be ready so we can integrate all commits at once. Alternatively, we can integrate t_memory_directive now and wait for other to be integrated next week. Up to you!

ryanemerson:
@vblagoje integrating all at once sounds better to me, so lets wait.

ryanemerson:
@vblagoje I've left the JIRA open as we will need to add the appropriate JSON etc to normal cache configuration pages eventually.

ryanemerson:
@vblagoje What's the plan with this PR? Are we revisiting it when time allows?

vblagoje:
@ryanemerson I'll review it next week and integrate it if server side DMR is ready. At the time I developed it - it was not.

vblagoje:
Postponing until DMR data is ready

PR #142, opened 2016-11-17
ISPN-6742 Console shows cache available and disabled at the same time

vblagoje:
Master only. Please verify and integrate @ryanemerson See https://issues.jboss.org/browse/ISPN-6742 fore more details

### DIFF
--- a/src/module/cache-container/caches/CachesCtrl.ts
+++ b/src/module/cache-container/caches/CachesCtrl.ts
@@ -6,9 +6,12 @@ import {AddCacheModalCtrl} from "./AddCacheModalCtrl";
 import IModalService = angular.ui.bootstrap.IModalService;
 import {ICacheContainer} from "../../../services/container/ICacheContainer";
 import {ITemplate} from "../../../services/container-config/ITemplate";
+import {CacheService} from "../../../services/cache/CacheService";
+import {IMap} from "../../../common/utils/IMap";
+import {isNullOrUndefined} from "../../../common/utils/Utils";
 
 export class CachesCtrl {
-  static $inject: string[] = ["$uibModal", "container", "caches", "templates"];
+  static $inject: string[] = ["$uibModal", "container", "caches", "templates", "cacheService"];
 
   traitCheckboxes: TraitCheckboxes = new TraitCheckboxes();
   typeCheckboxes: TypeCheckboxes = new TypeCheckboxes();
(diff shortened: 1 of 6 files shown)


vblagoje:
screen shot 2016-11-17 at 9 32 10 am

@pzapataf @ryanemerson if you have better UI suggestions for this use case - let me know.

ryanemerson:
@vblagoje In what circumstances is a cache UNAVAILABLE and enabled? I'm just thinking whether we need both of these flags, or could we just have one?

vblagoje:
Yeah, this screen sorta evolved and we could probably go with just one label. Thoughts @pzapataf @tristantarrant

vblagoje:
@ryanemerson I'll close this one as we need to discuss this it a bit more

vblagoje:
@pzapataf @ryanemerson I propose that until we have a robust solution that takes into account availability/status of the cache across the entire cluster - we pragmatically solve it now as well by fetching cache-availability cache attribute from the first server in the cluster (group). Such a solution is presented in this PR. Have a look.

ryanemerson:
@vblagoje, @pzapataf LGTM. If I don't hear any objections by the end of the day I'll integrate.

vblagoje:
@ryanemerson yes but with my proposal we'll get complaints about not actually addressing the core of the issue. We'll still have both AVAILABLE and disabled - which is ok if we can explain this in the documentation. I don't know myself anymore. Need a fresh look on Monday!

ryanemerson:
@vblagoje Any more thoughts on this?

vblagoje:
@ryanemerson I tend to agree with @pzapataf that we should keep two statuses separate. One is for the health of the cache (cache-availability) and the other is its external visibility (enabled/disabled). I admit this is a bit non-intuitive, as well all agreed, so until we come up with some better way to describe these states let's keep what we have. We need to convey this resolution to @Holmistr and @mgencur

Holmistr:
@vblagoje @ryanemerson  I understand your arguments and I'm fine with the proposed deal. Actually I'm thinking about some reasonable resolution. My main concern was that having those two label seems a bit non-intuitive, but I know that the statuses are not mutually exclusive. How about if we just changed the labels a bit, like prefix the status with information what it actually refers to? For example: we add string (or an icon/graphics) "Health: " + available/unavailable and (perhaps something like) "Activation:" (or "Exposed"/"Publicly available") + enabled/disabled.

If we made it obvious that the two labels refer to different properties of the cache, I think that solves the core problem - being non-intuitive. WDYT @vblagoje @ryanemerson @pzapataf  ?

vblagoje:
@ryanemerson please see this additional commit. Although it adjusts icons on container cache cards I think it will be a performance issue in cache container views where we have dozens or even hundreds of caches. If you are also concerned please remove this commit and integrate PR without it!

vblagoje:
@Holmistr I like your solution as well. Let's revisit it again once we have a bit more time!

ryanemerson:
Integrated afd786e6518fc188036a7dee18ab221acf1d1149, thanks @vblagoje

PR #143, opened 2016-11-17
ISPN-7206 Administration console - task history only shows last 10 tasks

vblagoje:
Master only. Do not integrate yet. @ryanemerson can you attempt to shrink select widget *not* to take entire page width? Give it a try before I call some more specialized CSS help

### DIFF
--- a/src/module/cache-container/tasks/TasksCtrl.ts
+++ b/src/module/cache-container/tasks/TasksCtrl.ts
@@ -16,6 +16,8 @@ export class TasksCtrl {
 
   history: IClusterEvent[] = [];
   runningTasks: ITaskStatus[] = [];
+  taskCount: number = 10;
+  taskCountOptions: number [] = [10, 50, 100];
 
   constructor(private $rootScope: IRootScopeService,
               private $uibModal: IModalService,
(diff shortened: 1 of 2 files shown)


ryanemerson:
@vblagoje I think the problem is that you have put the dropdown in the table header, which will mean that the dropdown is always the width of the table which is designed to take up the width of the page. To solve this problem I would putt the dropdown in a separate div to the table. 

Just a thought. A more flexible solution to this problem, would be to utilise [angular bootstrap's pagination](https://angular-ui.github.io/bootstrap/#/pagination) and only retrieve the  `x` most recent tasks on the first page, and then `x -> 2x` on the second page etc. The event log api supports this via the "since" field, so we retrieve `x` tasks and then when page 2 of the task history is set, we use the timestamp of task `x` as the date in the since field and again limit the total results to `x`. 

N.B. This is also an issue on the CacheContainer=>Config=>Tasks screen

vblagoje:
@ryanemerson I agree. I'll resolve this one as intended and then we'll open an Enhancement request to introduce pagination consistently across all pages where applicable.

vblagoje:
@ryanemerson Added a dropdown. Let's do pagination later as we discussed. Please verify @ryanemerson

ryanemerson:
@vblagoje I think you might have pushed the wrong commit ;)

vblagoje:
Right @ryanemerson . Something went wrong! I'll correct it now

vblagoje:
Ok, have a look now @ryanemerson

PR #144, opened 2016-11-18
ISPN-7214 Scope server restarts to server group

vblagoje:
Master only. See https://issues.jboss.org/browse/ISPN-7214 for more details. We need it for JDG-447


ryanemerson:
@vblagoje LGTM, but at the minute I can't get  the latest build of the server to restart without errors due to ISPN-7220, so unless there is a rush to get this in, I think we should wait so that we can properly test this.

vblagoje:
No that's cool @ryanemerson it can wait a few days. It is important we get it in because it resolves two product jiras

ryanemerson:
@vblagoje Integrated, ISPN-7220 has been fixed

PR #147, opened 2016-11-23
Disable minification on upstream.

ryanemerson:
We need to disable minification on upstream so that the console works on the next upstream release of Infinispan.


vblagoje:
Integrated. Thanks @ryanemerson

PR #148, opened 2016-11-24
Completed tasks are not shown in task history right away

vblagoje:
Master only. Trivial state reload on ui-sref transitions.


ryanemerson:
@vblagoje JDG PR https://github.com/infinispan/jdg-management-console/pull/46

PR #150, opened 2016-11-28
ISPN-7185 Confirmation dialog is not displayed when changing configur…

vblagoje:
Master only. Fixes JDG-561 blocker. Please review and integrate @ryanemerson

### DIFF
--- a/src/module/cache-templates/CacheTemplatesCtrl.ts
+++ b/src/module/cache-templates/CacheTemplatesCtrl.ts
@@ -64,16 +64,18 @@ export class CacheTemplatesCtrl extends AbstractConfigurationCtrl {
   }
 
   updateTemplate(): void {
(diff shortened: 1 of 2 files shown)


ryanemerson:
@vblagoje Should this also apply to changing cache configuration? At the moment this only addresses templates.

vblagoje:
@ryanemerson lets ask Roman Macor

rmacor:
@vblagoje @ryanemerson This should apply to cache configuration as well, actually, in the previous version it didn't apply to templates, but I suppose it's better if it's in both places - looks more consistent. WDYT?

ryanemerson:
@rmacor @vblagoje Personally I think it should be both, as the consequences of changing  a configuration and a template are similar, so it makes sense that there should be the same number of hoops to jump through for each action.

vblagoje:
Updated @ryanemerson Please review and integrate!

PR #154, opened 2016-11-30
ISPN-7254 Accessing content without needed permissions should display error message

vblagoje:
Master only. Error dialog is raised. Upon clicking OK admin is taken to containers page. See https://issues.jboss.org/browse/ISPN-7254 for more details

### DIFF
--- a/src/ManagementConsole.ts
+++ b/src/ManagementConsole.ts
@@ -33,6 +33,8 @@ import IAngularEvent = angular.IAngularEvent;
 import {FileModelDirective} from "./components/directives/FileModelDirective";
 import {VertilizeContainerDirective} from "./components/directives/VertilizeContainerDirective";
 import {VertilizeDirective} from "./components/directives/VertilizeDirective";
+import {openErrorModal} from "./common/dialogs/Modals";
+import IModalService = angular.ui.bootstrap.IModalService;
 
 const App: ng.IAngularStatic = angular;
 
@@ -142,10 +144,12 @@ module.run(($templateCache: ITemplateCacheService) => {
 
 // TODO remove if only used for debugging
 // @ngInject
-module.run(($rootScope: IRootScopeService) => {
+module.run(($rootScope: IRootScopeService, $uibModal: IModalService, $state: IStateService) => {
   $rootScope.$on("$stateChangeError", (event: IAngularEvent, toState: any, toParams: any, fromState: any, fromParams: any, error: any) => {
-      alert("error: " + error);
+    openErrorModal($uibModal, error).result.then(() => {
+      $state.go("containers", null, {reload: true});
     });
+  });
 });
 
 export {App};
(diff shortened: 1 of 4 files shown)


src/services/schemas/SchemasService.ts
```diff
@@ -25,7 +25,7 @@ export class SchemaService {
     let deferred: ng.IDeferred<ISchemaDefinition> = this.$q.defer<ISchemaDefinition>();
     this.findTarget(container.serverGroup)
       .then(coordinator => {
-          return this.dmrService.executePost({
```
ryanemerson: We need this return, as this means that the result of the promise `this.dmrService.executePost(...` is passed to L35's then clause. It does not mean that the overarching function `getProtoSchema` returns this result.

src/services/schemas/SchemasService.ts
```diff
@@ -68,7 +71,7 @@ export class SchemaService {
     let deferred: ng.IDeferred<void> = this.$q.defer<void>();
     this.findTarget(container.serverGroup)
       .then(coordinator => {
-        return this.dmrService.executePost({
+        this.dmrService.executePost({
```
ryanemerson: We need this return as it ensures that the result of the executePost is passed to the the chained `then.` call

src/services/schemas/SchemasService.ts
```diff
@@ -39,12 +39,15 @@ export class SchemaService {
...
           operation: "get-proto-schema-names"
-        }).then(names => deferred.resolve(names));
+        }).then(
+          (names) => deferred.resolve(names),
+          (error) => deferred.reject(error));
       },
-      error => deferred.reject(error));
+      (error) => deferred.reject(error)
+    );
     return deferred.promise;
```
ryanemerson: The original code is definitely wrong, but I believe that the second `.then` should be chained after the first, see below:

```typescript
registerProtoSchema(container: ICacheContainer, schema: ISchemaDefinition): ng.IPromise<void> {
    let deferred: ng.IDeferred<void> = this.$q.defer<void>();
    this.findTarget(container.serverGroup)
      .then(coordinator => {
          return this.dmrService.executePost({
            address: this.getContainerAddress(container.name, coordinator),
            operation: "register-proto-schemas",
            "file-names": [schema.fileName],
            "file-contents": [schema.fileContents]
          });
        },
        error => deferred.reject(error))
      .then(() => deferred.resolve(), error => deferred.reject(error));
    return deferred.promise;
  }
```

src/ManagementConsole.ts
```diff
@@ -142,10 +144,12 @@ module.run(($templateCache: ITemplateCacheService) => {
 
 // TODO remove if only used for debugging
```
ryanemerson: Lets remove this TODO

src/module/cache-container/config/ContainerConfig.ts
```diff
@@ -11,6 +11,7 @@ import {ContainerConfigService} from "../../../services/container-config/Contain
 import {TasksCtrl} from "./tasks/TasksCtrl";
 import {SecurityCtrl} from "./security/SecurityCtrl";
 import IQService = angular.IQService;
+import {ContainerService} from "../../../services/container/ContainerService";
```
ryanemerson: Unnecessary import added.

src/services/schemas/SchemasService.ts
vblagoje: +1

vblagoje:
@ryanemerson updated. Please check again and if all good integrate and issue PR for the product side as well

ryanemerson:
Integrated cf411575cc5db0586890867fd473a4c3d77708f7, thanks @vblagoje

PR #155, opened 2016-11-30
ISPN-7261: Modal dialogs totally obscure background.

ryanemerson:
https://issues.jboss.org/browse/ISPN-7261


ryanemerson:
@vblagoje The problem was caused by us importing the bootstrap css and patternfly's css. This caused some of the required css to be overridden as patternfly already pulls in bootstrap's css.

I have tested this using `gulp serve:dist` and by generating the jar and manually putting it on the server.

vblagoje:
Verified and integrated with 7671eb5 Thanks @ryanemerson

PR #156, opened 2016-11-30
ISPN-7251 Changing configuration in standalone mode should not prompt restart dialog

vblagoje:
Master only. See https://issues.jboss.org/browse/ISPN-7251 for more details


src/module/cache-container/config/thread-pools/ThreadPoolsCtrl.ts
```diff
@@ -29,7 +29,7 @@ export class ThreadPoolsCtrl extends AbstractConfigurationCtrl {
     this.containerConfigService.saveThreadPools(this.container, this.threadPools)
       .then(() => {
         if (this.isRestartRequired()) {
-          if (this.launchType.isStandaloneLocalMode()) {
+          if (this.launchType.isStandaloneMode()) {
```
vblagoje: @ryanemerson having just if else switch here is ok because we want to have one type of dialog for both standalone types (clustered and non-clustered) and another for domain mode!

PR #157, opened 2016-11-30
ISPN-7247: Administration console - changing any cache configuration …

ryanemerson:
…adds security tag to the configuration.

https://issues.jboss.org/browse/ISPN-7247

@vblagoje This prevents the <security> tags being added to a configuration when security isn't enabled and it allows config to be updated etc and the server restarted without errors. One problem at the minute though, is that if security is enabled for the container and a cache config is updated, then the following is added to the config:

```xml
<security>
    <authorization enabled="false" roles=""/>
</security>
```

^ Doesn't result in any errors (AFAIK), but it would cause an issue if the user disables container security in the xml and doesn't update their cache configs. Review and let me know what you think.

### DIFF
--- a/src/components/cache-configuration/view/distributed-cache.html
+++ b/src/components/cache-configuration/view/distributed-cache.html
@@ -45,7 +45,7 @@
     </uib-tab>
 
     <uib-tab heading="Security">
-      <cache-security container="$ctrl.container" data="$ctrl.data.security.SECURITY.authorization.AUTHORIZATION"
+      <cache-security container="$ctrl.container" data="$ctrl.data.security.SECURITY"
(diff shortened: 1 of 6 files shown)


vblagoje:
@ryanemerson it works. I'll integrate it as I did not notice any issues. We should, however, consider additional ways to check dirty fields from ConfigService

ryanemerson:
@vblagoje 100%. I think we could make that code a lot more flexible in general so that it is easier to resolve such issues

vblagoje:
Integrated with 1ffa6010d3d90b713fcb3ef671e07aa2d4e50f9d Thanks @ryanemerson

PR #158, opened 2016-11-30
Cache-availability attribute does not exist for Local Caches.

ryanemerson:
@vblagoje cache-availability attribute doesn't exist for local caches in either Standalone or Domain mode. My understanding is that if server containing the local cache is available, then the cache itself is AVAILABLE. Without this fix, it's not possible to resolve ISPN-7251 as I was not able to access the cache configuration screens due to the errors being thrown.


vblagoje:
Resolved with e2a67cd. Thanks @ryanemerson

PR #159, opened 2016-12-01
ISPN-7215: Upload of deployment artifact doesn't work.

ryanemerson:
https://issues.jboss.org/browse/ISPN-7215

~~@vblagoje Do you think we can get this into the release?~~ It has already began

This works with both `serve:dist` and deploying a generated jar into the server (make sure you hit clear cache and hard reload when trying the server for the first time, that wasted 10 minutes of me having to diff the jar and dist folder contents :/)


vblagoje:
Amazing fix. Integrated with f7b4ea1 Thanks a bunch @ryanemerson

PR #160, opened 2016-12-02
ISPN-7264 BULK_READ permission is not set in identity/role (workaround)

vblagoje:
Master only. Workaround until we solve the mystery of BULK_READ permission for get-proto-schema-names DMR call


ryanemerson:
@vblagoje I haven't had a chance to test this yet, but can you briefly explain the rational behind this approach? I'm confused to why this would work

vblagoje:
@ryanemerson when admin clicks on cache container configuration we do forward/redirect on schemas tab. That's the problematic DMR call causing authorization exception to be raised. If we make transport default then at least we can allow admin to see other tabs and use them! Makes sense?

PR #161, opened 2016-12-02
Standalone page refresh causes the whole app to die.

ryanemerson:
https://issues.jboss.org/browse/ISPN-7273


vblagoje:
Integrated with 8d21d404bed3d71fab1bcfa69160f39f998d9692 thanks @ryanemerson

PR #164, opened 2016-12-07
ISPN-7242 Administration console does not work on IE

vblagoje:
Master only. @ryanemerson do i need to change jspm.conf.js

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,7 +9,7 @@
 <body>
 
 <div>
-  <div ui-view name="main"/>
+  <ui-view name="main"></ui-view>
 </div>
 
 <script src="vendor/jspm_packages/system.js"></script>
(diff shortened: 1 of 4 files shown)


src/jspm.conf.js
```diff
@@ -205,9 +205,21 @@ System.config({
       "crypto": "github:jspm/nodelibs-crypto@0.1.0",
       "url": "github:jspm/nodelibs-url@0.1.0"
     },
+    "npm:babel-polyfill@6.16.0": {
```
ryanemerson: @vblagoje I think we need all of them as the changes further down are defining what `babel-polyfill` depends on.

src/index.html
```diff
@@ -9,7 +9,7 @@
 <body>
 
 <div>
-  <div ui-view name="main"/>
+  <ui-view name="main"></ui-view>
```
ryanemerson: @vblagoje We need to make this change to `dist/index.html` as well

vblagoje:
This PR is obsololete as (ISPN-7310: Make Console Use Digest Authentication) 948b1bd502b6e19f7d56305ed0b911ab2c93d077 resolves this issue as well

PR #168, opened 2016-12-14
ISPN-7242: Update dist index.html.

ryanemerson:
Necessary for IE 10 when we package the console.


ryanemerson:
@vblagoje This will also be necessary on JDG

PR #169, opened 2016-12-15
ISPN-7247 changing any cache configuration adds security tag to the configuration (addendum)

vblagoje:
Master only. Do not blindly integrate. This is a working proposal to be considered

### DIFF
--- a/src/components/cache-security/CacheSecurityCtrl.ts
+++ b/src/components/cache-security/CacheSecurityCtrl.ts
@@ -9,7 +9,7 @@ export class CacheSecurityCtrl implements IConfigurationCallback {
   static $inject: string[] = ["containerService"];
 
   data: any;
-  auth: any;
+  auth: any = {"is-dirty": false};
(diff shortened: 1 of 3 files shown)


ryanemerson:
@vblagoje It works for me :). I'm happy with the use of "is-dirty" for extra control in the service as we should be able to apply this key word for other config aspects when required. Let me know if you wish to add anything to this, otherwise I'll integrate by the end of the day

vblagoje:
@ryanemerson I don't mind. I wanted to see with you if we should be a bit more systematic immediately and use "is-dirty" across the board? Maybe this gradual introduction on per-needed basis is ok too. Thoughts?

ryanemerson:
@vblagoje I think a gradual introduction will be less likely to cause regressions for now and then we can revisit later if/when required.

vblagoje:
Ok deal @ryanemerson Please check all variations of security handling. I did as well but after you check it as well I know we are good to go!

PR #170, opened 2017-01-04
ISPN-7330 Cannot edit configuration twice without refresh

vblagoje:
Master only. @ryanemerson I will add "Server restarting dialog" as well a bit later. But for now, I think this is an appropriate solution for ISPN-7330 Please review and comment. Thank you


ryanemerson:
@vblagoje I tested this and merged, but then I realised that there was missing semi-colons on several lines (the coffee must not have kicked in yet), hence the need for this commit 3fa214c5db2bfead202873b8e066bc4e37d68f35. The JDG PR contains these fixes as a single commit.

PR #173, opened 2017-01-10
ISPN-7348: Cannot create new cache with template containing backup si…

ryanemerson:
…tes.

https://issues.jboss.org/browse/ISPN-7348

@vblagoje I know we don't currently allow for the configuration of backup sites [ISPN-6529](https://issues.jboss.org/browse/ISPN-6529), however if we are creating a config from an existing template, it makes sense for that underlying template to be copied in full to the new cache config, even if the user cannot modify these values in the console itself.

### DIFF
--- a/src/services/cache-config/CacheConfigService.ts
+++ b/src/services/cache-config/CacheConfigService.ts
@@ -296,7 +296,7 @@ export class CacheConfigService {
         this.createHelper(builder, address.concat("partition-handling", "PARTITION_HANDLING"), config["partition-handling"]);
         this.createHelper(builder, address.concat("transaction", "TRANSACTION"), config.transaction);
         this.createHelper(builder, address.concat("state-transfer", "STATE_TRANSFER"), config["state-transfer"]);
-        this.createHelper(builder, address.concat("backup", "BACKUP"), config.backup);
+        this.createBackupNodes(builder, address.concat("backup"), config.backup);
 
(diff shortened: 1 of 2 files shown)


src/services/cache-config/CacheConfigService.ts
```diff
@@ -374,7 +386,6 @@ export class CacheConfigService {
     this.updateHelper(builder, address.concat("partition-handling", "PARTITION_HANDLING"), config["partition-handling"]);
     this.updateHelper(builder, address.concat("transaction", "TRANSACTION"), config.transaction);
     this.updateHelper(builder, address.concat("state-transfer", "STATE_TRANSFER"), config["state-transfer"]);
-    this.updateHelper(builder, address.concat("backup", "BACKUP"), config.backup);
```
ryanemerson: @vblagoje We don't allow backups to be updated in the console, so it doesn't make sense to add the extra dmr step if we know nothing will have changed. Or am I missing something?

vblagoje: Right right, I got it now that I focused on the overall issues rather than just the code changes. All good @ryanemerson

ryanemerson:
@vblagoje Updated to solve the "write-behind" issue, this latest commit is the same logic as ISPN-7219

vblagoje:
Ok, verified, squashed, tested and integrated with 7fde3d0 Thanks @ryanemerson

PR #174, opened 2017-01-10
ISPN-6988 Create UI for off-heap configuration

vblagoje:
Master only. Please review but don't integrate yet @ryanemerson

### DIFF
--- a/src/components/memory/MemoryCtrl.ts
+++ b/src/components/memory/MemoryCtrl.ts
@@ -35,11 +35,10 @@ export class MemoryCtrl implements IConfigurationCallback {
         this.data.initialType = field;
       }
       this.data["is-new-node"] = false;
-    }
-    else {
-      this.data = {};
-      this.setMemoryType("NONE");
-      this.data["is-new-node"] = true;
+    } else {
+        this.data = {};
+        this.setMemoryType("NONE");
+        this.data["is-new-node"] = true;
     }
     this.prevData = {};
     this.cleanMetadata();
@@ -67,7 +66,7 @@ export class MemoryCtrl implements IConfigurationCallback {
     let fields: string [] = [];
     for (let field in data) {
       if (MEMORY_TYPES.indexOf(field) > -1) {
-        fields.push(field)
+        fields.push(field);
       }
     }
     return fields;
--- a/src/components/memory/MemoryCtrl.ts
+++ b/src/components/memory/MemoryCtrl.ts
@@ -1,0 +1,117 @@
+import {IConfigurationCallback} from "../../common/configuration/IConfigurationCallback";
+import {isNotNullOrUndefined, deepGet} from "../../common/utils/Utils";
+import {
+  isFieldValueModified,
+  fieldChangeRequiresRestart,
+  makeFieldClean,
+  makeFieldDirty
+} from "../../common/configuration/ConfigUtil";
+
+export const MEMORY_TYPES: string [] = [ "NONE", "BINARY", "OBJECT", "OFF-HEAP"];
+
+export class MemoryCtrl implements IConfigurationCallback {
+
+  data: any;
+  meta: any;
+  memoryTypeData: any;
+  memoryTypeMeta: any;
+  readOnly: boolean;
+  configCallbacks: IConfigurationCallback[];
+  memoryTypes: string [];
+  fields: string[] = [];
+  allfields: string [] = [];
+  prevData: any;
+  type: string = "";
+
+  constructor() {
+    this.fields.length = 0;
+    this.memoryTypes = MEMORY_TYPES;
+    if (isNotNullOrUndefined(this.configCallbacks)) {
+      this.configCallbacks.push(this);
+    }
+    if (isNotNullOrUndefined(this.data)) {
+      for (let field of this.filterFields(this.data)) {
+        this.setMemoryType(field);
+        this.data.initialType = field;
+      }
+      this.data["is-new-node"] = false;
+    }
+    else {
+      this.data = {};
+      this.setMemoryType("NONE");
+      this.data["is-new-node"] = true;
+    }
+    this.prevData = {};
+    this.cleanMetadata();
+  }
+
+  setMemoryType(newType: string): void {
+    if (isNotNullOrUndefined(newType)) {
+      this.fields.length = 0;
+      this.type = newType;
+      this.memoryTypeData = isNotNullOrUndefined(this.data[newType]) ? this.data[newType] : {};
+      this.memoryTypeMeta = (newType !== "NONE") ? angular.copy(deepGet(this.meta, "children.memory.model-description." + newType + ".attributes")) : {};
+      for (var field in this.memoryTypeMeta) {
+        this.fields.push(field);
+      }
+      this.data[newType] = this.memoryTypeData;
+      this.memoryTypeData.type = newType;
+      this.memoryTypeMeta.type = {description: "Memory management type"};
+
+      this.allfields = angular.copy(this.fields);
+      this.allfields.push("type");
+    }
+  }
+
+  filterFields(data: any): string [] {
+    let fields: string [] = [];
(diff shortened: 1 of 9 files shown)


src/components/memory/view/memory.html
```diff
@@ -0,0 +1,27 @@
...
+  <div class="row">
+    <div class="panel panel-default">
+      <div class="panel-heading">
+        <h3 class="panel-title">Memory</h3>
+      </div>
+      <div class="panel-body">
+        <div class="form-horizontal" ng-if="!$ctrl.isMemoryDefined()">
+          <div class="form-group">
+            <label class="col-md-8 control-label">Memory management has not been defined. Click
+              <button ng-click = "$ctrl.createDefaultMemory()"> here</button> to define it</label>
```
ryanemerson: I do like this message, however what does a user do if they want to remove the memory configuration on an existing cache/template? With stores/loaders they can simply select the No option in the combo box and the previously configured store/loader is removed on update.  Obviously we would need to add this logic to the CacheConfigService as well.

src/components/memory/MemoryCtrl.ts
```diff
@@ -0,0 +1,127 @@
...
+  getStyle(field: string): string {
+    return isNotNullOrUndefined(this.memoryTypeMeta[field]) ? this.memoryTypeMeta[field].style : "";
+  }
+
+  private undoFieldChange(field: string): void {
+    this.memoryTypeData[field] = this.prevData[field];
+    makeFieldClean(this.memoryTypeMeta[field]);
+  }
+
+  private makeMetaForType(): any {
```
ryanemerson: Do we really need this method? It's only called once and it only returns a one liner

src/services/cache-config/CacheConfigService.ts
```diff
@@ -535,6 +557,14 @@ export class CacheConfigService {
     return op;
   }
 
+  private createRemoveOperation(address: string[]): IDmrRequest {
+    let op: IDmrRequest = {
+      address: address,
+      operation: "remove"
+    };
+    return op;
```
ryanemerson: No need for `op` variable, we can just return the object directly.

src/services/cache-config/CacheConfigService.ts
```diff
@@ -387,6 +398,17 @@ export class CacheConfigService {
     return this.dmrService.executePost(builder.build());
   }
 
+  private getMemoryType(memory: any): string {
+    let memoryType: string = "";
+    for (var fieldName in memory) {
+      // there should be only one object BINARY, OBJECT, or OFF-HEAP
+      if (isObject(memory[fieldName])) {
+        memoryType = fieldName;
```
ryanemerson: If there is only one object, can we not just return `fieldName` here and remove the `memoryType` variable?

src/components/memory/MemoryCtrl.ts
```diff
@@ -0,0 +1,127 @@
...
+  {
+    value: "BINARY"
+  },
+  {
+    value: "OBJECT"
+  },
+  {
+    value: "OFF-HEAP"
+  }
+];
```
vblagoje: Right. Spot on

src/components/memory/MemoryCtrl.ts
```diff
@@ -0,0 +1,127 @@
...
+      this.memoryTypeMeta = angular.copy(deepGet(this.meta, "children.memory.model-description." + newType + ".attributes"));
+      for (var field in this.memoryTypeMeta) {
+        this.fields.push(field);
+      }
+      if (isNotNullOrUndefined(oldType) && oldType !== newType) {
+        delete this.data[oldType];
+        this.memoryTypeData.removeType = oldType;
+      }
+      this.data[newType] = this.memoryTypeData;
+      this.memoryTypeData.type = newType;
```
vblagoje: Great catch. Looking into it

src/components/memory/MemoryCtrl.ts
vblagoje: +1

src/services/cache-config/CacheConfigService.ts
```diff
@@ -367,7 +367,18 @@ export class CacheConfigService {
...
+      isNotNullOrUndefined(config.memory[memoryType].removeType);
+    let firstTimeAddingMemory:boolean = isNotNullOrUndefined(config.memory[memoryType]) && config.memory["is-new-node"];
+    if (switchingMemoryType) {
+      builder.add(this.createRemoveOperation(address.concat("memory", config.memory[memoryType].removeType)));
+      this.createHelper(builder, address.concat("memory", memoryType), config.memory);
+    } else if (firstTimeAddingMemory) {
+      this.createHelper(builder, address.concat("memory", memoryType), config.memory);
+    } else {
+      this.updateHelper(builder, address.concat("memory", memoryType), config.memory);
+    }
```
vblagoje: +1

src/services/cache-config/CacheConfigService.ts
vblagoje: +1

src/services/cache-config/CacheConfigService.ts
vblagoje: Yes we can!

src/components/memory/view/memory.html
vblagoje: Ok, makes sense! Do you prefer no option instead of this message? I am breaking UI consistency with this message.

vblagoje:
Thanks will address these today and make a new version for tomorrow! Thanks @ryanemerson

src/components/memory/view/memory.html
ryanemerson: I guess it makes sense to keep it consistent with the stores/loaders. Long term maybe we should come up with a generic option for removing an element from the config, something like a red cross at the top right of the individual screens. We could then apply this to indexing etc as well, but obviously not a priority with all the other stuff we have on.

vblagoje:
@ryanemerson give it another look. It should be much closer to what we want now!

ryanemerson:
@vblagoje I have found several errors:

- On update an existing cache, the server returns:
```javascript
"WFLYCTL0216: Management resource '[
    ("profile" => "clustered"),
    ("subsystem" => "datagrid-infinispan"),
    ("cache-container" => "clustered"),
    ("configurations" => "CONFIGURATIONS"),
    ("distributed-cache-configuration" => "default"),
    ("memory" => "OFF-HEAP")
]' not found"
```

I think this is caused by "is-new-node" is not being set to true on the memory node.

- Updating existing template: memory already defined but size isn't, error when adding size value and pressing update.
```javascript
WFLYCTL0030: No resource definition is registered for address [
    ("profile" => "clustered"),
    ("subsystem" => "datagrid-infinispan"),
    ("cache-container" => "clustered"),
    ("configurations" => "CONFIGURATIONS"),
    ("replicated-cache-configuration" => "replicated"),
    ("memory" => "undefined")
]
```

- When loading create new template page using existing template which has memory defined we get the following javascript error:

```javascript
angular.js:13923 TypeError: Cannot create property 'type' on boolean 'true'
    at MemoryCtrl.setMemoryType (MemoryCtrl.ts:62)
    at new MemoryCtrl (MemoryCtrl.ts:35)
    at Object.invoke (angular.js:4721)
    at $controllerInit (angular.js:10357)
    at nodeLinkFn (angular.js:9266)
    at eval (angular.js:9676)
    at processQueue (angular.js:16386)
    at eval (angular.js:16402)
```

I Can still create new template, however the Memory object isn't copied from the base template to the newly create template.


- Finally, there are a few new failures when running `gulp check` that should be fixed.

vblagoje:
@ryanemerson I checked all the conditions you mention. If it seems ok to you as well (and please test it thoroughly), let's integrate this one for the next community release.

ryanemerson:
@vblagoje Integrated e0ab93b7d6eaca1f3d5cf3264d86a305fe337940, thanks!

PR #177, opened 2017-01-17
ISPN-7244 Creating new cache should only show valid cache types

vblagoje:
Master only. See https://issues.jboss.org/browse/ISPN-7244 Please verify local mode and domain mode to test this functionality @ryanemerson

### DIFF
--- a/src/module/cache-container/caches/AddCacheModalCtrl.ts
+++ b/src/module/cache-container/caches/AddCacheModalCtrl.ts
@@ -4,10 +4,10 @@ import {ICacheContainer} from "../../../services/container/ICacheContainer";
 import {CacheService} from "../../../services/cache/CacheService";
 import {openErrorModal, openConfirmationModal} from "../../../common/dialogs/Modals";
 import IModalService = angular.ui.bootstrap.IModalService;
-import {LaunchTypeService} from "../../../services/launchtype/LaunchTypeService";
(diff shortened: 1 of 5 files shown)


ryanemerson:
@vblagoje We should define the behaviour of default cache types for standalone/domain all in the CacheConfigService so that the logic is centralised. I believe the following actions are required:

1. Replace `CACHE_TYPES` in `CacheConfigService` with  a `DOMAIN_CACHE_TYPES` and `STANDALONE_CACHE_TYPES` arrays. 

2. As the `CacheConfigService` already injects the launch service, create the following methods in `CacheConfigService`:

    - `getAllCacheTypes(); // return appropriate array based upon launch type`
    - `getDefaultCacheType() // return the first index of the array returned by ^ method'
    - `filterTemplates(template[]) // filter passed templates based upon getAllCacheTypes`

The advantage of ^ approach is that all of the logic for default/valid template/cache types is dictated by the two arrays. This will also reduce the size of the resolve logic for `CacheTemplates` and `Caches`.

vblagoje:
Great ideas and insights @ryanemerson Will follow your suggestions!

vblagoje:
Thanks for these awesome insights @ryanemerson Onto id generator now!

ryanemerson:
@vblagoje Integrated c592d7f4ef06a63e2ff11831f98b4a67b343cc54, thanks

PR #179, opened 2017-01-18
ISPN-5941: Ensure testability of management console. (WIP)

ryanemerson:
https://issues.jboss.org/browse/ISPN-5941

@vblagoje I have created a directive that generates id attributes for divs/elements based upon other attributes, the current page and provided values in a `id-tag` attribute.  The directive is declared as `id-generator`.

So far I have applied this to all of the config screens, the `containers` page and the `caches` page. Let me know what you think of the approach.

NOTE: As we require DOM manipulation it was not possible to use a component, hence a directive was used.


ryanemerson:
@vblagoje Just the modals to go now, but I'll refrain from implementing those changes there until I hear back from you as updating the html is a pretty tedious task.

vblagoje:
@ryanemerson I like the approach. A few questions but I think you are already aware of this as the PR is WIP. You will somehow externalize getID function rather that re-implement it is in each component, right? Why do we need id-tag attribute? Can't we pass whatever parameters we need to id-generator?

ryanemerson:
@vblagoje You're right, I can definitely reduce the repetition of the getId function declarations. 

I need to look more into pass parameters to a directive when it is declared as an attribute, but you're right that would make more sense.

ryanemerson:
@vblagoje I managed to remove the need for `id-tag`, good idea. I just need to update the modals now ..

ryanemerson:
@vblagoje I have updated the remaining html, let me know what you think.

ryanemerson:
@Holmistr @rmacor Can you please review this PR and make sure it provides the element ids you require and that the generate ids are ok. You will need to pull the code and run it locally to see what the generated id's look like. The way I generate the ids means that they should only change if the value passed to id-generator in the html changes or if we change the name of routes (which we don't plan on doing unless absolutely necessary).

vblagoje:
@ryanemerson this is some amazing piece of work man! So now basically we only use id-generator and if type and value are present in the html element we use them as well. If not, then they are not used. Is that correct? Will run this actual code now and see observe the actual generation of ids!

ryanemerson:
@vblagoje Yeah that's it! All of the logic for actually generating the id is [here](https://github.com/ryanemerson/infinispan-management-console/blob/b0e100c2aa67414279134908436271dc01fe3e5a/src/components/directives/IdGeneratorDirective.ts#L35-L48l).

vblagoje:
Still going through but so far I found that individual caches on containers/clustered/clustered/caches page likely need fine grained id generation as right now each cache has the same id. How would the script identify a particular cache and click on it? Maybe @Holmistr and @rmacor know more about this requirement?

ryanemerson:
@vblagoje A cache must have a unique name within a container, so I guess it's probably sufficient to just incorporate the cache name within the div right?

vblagoje:
@ryanemerson yes within the container. I would just add it as value attribute, right? And in all other cases where we need to - we can use the same approach.

ryanemerson:
@vblagoje I would pass it as part of the id-generator="<cache_name>" along with anything else. No need to introduce the value attribute in this case as it may confuse things later on.

vblagoje:
Right, right. Forgot you can actually pass runtime parameters. This is just awesome :-)

rmacor:
@ryanemerson it looks great, the generated IDs are suitable, just a few comments:
- The IDs are missing in some places, I've noticed Cluster tab for example (values for nodes, hosts) and also the buttons that are disabled don't have IDs
- You are removing some IDs (#cacheCards, myModalLabel) without creating new ones
- Since you are also changing the existing IDs we will need to update the test suite, but we agreed that's doable as long as the IDs won't change again in the future

ryanemerson:
@rmacor Thanks for the feedback. I removed the myModalLabel as it was for non-generated content. But I can easily replace these with the new scheme if they are still valuable.  

Unfortunately the changes to the test suite are inevitable, but we don't plan on changing the ids once in place :)

ryanemerson:
@vblagoje  I have updated based upon @rmacor's feedback and squashed the commits. Unless you, @rmacor  or @Holmistr have anything else to add, this is ready to be merged imo.

rmacor:
I have no other comments, thanks

vblagoje:
@rmacor and @ryanemerson integrated with 7057bc3 Great work and all around participation guys

PR #181, opened 2017-01-26
ISPN-7353 Additional information in new script dialog is not displayed

vblagoje:
Trivial but verify please @ryanemerson I tried Chrome and Firefox


ryanemerson:
@vblagoje integrated, thanks

PR #182, opened 2017-01-30
Make all modal tooltips left-bottom positioned

vblagoje:
Do not integrate @ryanemerson . This is just a proposal for you to consider!

### DIFF
--- a/src/module/cache-container/config/security/view/add-role-modal.html
+++ b/src/module/cache-container/config/security/view/add-role-modal.html
@@ -13,8 +13,8 @@
           <input id-generator="role.name" type="text" class="form-control" ng-model="ctrl.role.name" ng-readonly="!ctrl.isNewRole">
         </div>
         <div class="col-md-2 pull-left" style="padding-left: 0">
-          <i class="fa fa-info-circle ng-scope" tooltip="The security role name" tooltip-trigger="" tooltip-animation="true"
-             tooltip-placement="top"></i>
+          <i class="fa fa-info-circle ng-scope" tooltip="The security role name"
+             tooltip-placement="left-bottom"></i>
         </div>
       </div>
       <div class="form-group">
--- a/src/module/cache-container/caches/view/add-cache-modal.html
+++ b/src/module/cache-container/caches/view/add-cache-modal.html
@@ -18,8 +18,8 @@
                     <input type="text" class="form-control" ng-model="ctrl.newCacheName" id-generator="create.name" required>
                 </div>
                 <div class="col-md-2 pull-left" style="padding-left: 0">
-                    <i class="fa fa-info-circle ng-scope" uib-tooltip="The name of the cache to be created" tooltip-trigger=""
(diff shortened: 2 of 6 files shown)


src/module/cache-container/config/security/view/add-role-modal.html
```diff
@@ -13,8 +13,8 @@ <h4 class="modal-title" id-generator="add_role_modal.title" translate>PERMISSION
           <input id-generator="role.name" type="text" class="form-control" ng-model="ctrl.role.name" ng-readonly="!ctrl.isNewRole">
         </div>
         <div class="col-md-2 pull-left" style="padding-left: 0">
-          <i class="fa fa-info-circle ng-scope" tooltip="The security role name" tooltip-trigger="" tooltip-animation="true"
-             tooltip-placement="top"></i>
+          <i class="fa fa-info-circle ng-scope" tooltip="The security role name"
+             tooltip-placement="left-bottom"></i>
```
ryanemerson: (Applies to master as well) This tooltip doesn't display as it still uses `tooltip=` when it should be `uib-tooltip=`.

ryanemerson:
@vblagoje I have nothing against the idea, however you still need to update the following files:

src/components/forms/field-info/view/field-info.html
src/components/cache-security/view/cache-security.html

Is there a reason for the change? Is having top aligned tooltips causing issues?

src/module/cache-container/config/security/view/add-role-modal.html
ryanemerson: Based upon a quick grep, this also applies to `src/components/cache-security/view/cache-security.html`

vblagoje:
@ryanemerson I wanted to have tooltips in modals (not field info and security) align left as we have made a few of them already align left in other modals (a commit or so ago). This way we stay consistent. Ideally we could make this a setting in a config. Will explore this a bit more after ISPN-7425 is done

ryanemerson:
@vblagoje Yeah having it configurable would be cool. A thought for the future, we could configure UI preferences like this via URL parameters that are applicable to all pages. As we use a base url that all pages inherit from, we would only have to write the resolution code once and then we could just inject a "UIProperties" object into the relevant controllers and the templates would call something like `$ctrl.uiProps.getTooltipAlignment()`. If no url params are specified, then we simply use default values.

vblagoje:
@ryanemerson I think I know what you mean but how would you store these UI preferences? One global default and then each page inherits these and overwrites them? How would $ctrl.uiProps.getTooltipAlignment() resolve the value? We need to write this UIProps class to do the value resolution!

ryanemerson:
@vblagoje You would store them in an object, which is then injected into controllers where required. For every page load, you a new instance of this object is created in the resolve method of the root state and it can then be injected as required. 

We could create a Immutable UIProperties class which stores all of the values and exposes them via getter methods.This class could then provide a static create method, which takes a $stateParams object as its only arg and creates a new UIProperties instance using the current URL parameters to initialize the instance's values.

PR #184, opened 2017-02-03
ISPN-7425 Enable admins to delete configuration subsections

vblagoje:
Master only. @ryanemerson this is just a preview so you can see if I am going in a right direction. Will complete it for Monday if it is. Have a look

### DIFF
--- a/src/components/configuration-section/ConfigurationSectionCtrl.ts
+++ b/src/components/configuration-section/ConfigurationSectionCtrl.ts
@@ -19,6 +19,7 @@ export class ConfigurationSectionCtrl implements IConfigurationCallback {
   configCallbacks: IConfigurationCallback[];
   removable: boolean;
   placeholders: any;
+  loadedWithData: boolean;
 
   constructor() {
     if (isNullOrUndefined(this.data)) {
@@ -28,7 +29,9 @@ export class ConfigurationSectionCtrl implements IConfigurationCallback {
       this.configCallbacks.push(this);
     }
     this.prevData = {};
-    this.data["is-new-node"] = !this.hasAnyFieldPreviousData();
+    let hasFieldsWithData: boolean = this.hasAnyFieldPreviousData();
+    this.loadedWithData = hasFieldsWithData;
+    this.data["is-new-node"] = !hasFieldsWithData;
     this.cleanMetadata();
     this.createPlaceholders();
   }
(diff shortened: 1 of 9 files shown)


src/components/configuration-section/ConfigurationSectionCtrl.ts
```diff
@@ -71,6 +72,47 @@ export class ConfigurationSectionCtrl implements IConfigurationCallback {
...
+    });
+  }
+
+  isRemovable(): boolean {
+    if (isNotNullOrUndefined(this.removable)) {
+      return this.removable;
+    } else {
+      return false;
+    }
+  }
```
ryanemerson: This can be made a one liner: 
```typescript
isRemovable: Function = () => isNotNullOrUndefined(this.removable) ? this.removable : false;
```

src/components/configuration-section/ConfigurationSectionCtrl.ts
```diff
@@ -71,6 +72,47 @@ export class ConfigurationSectionCtrl implements IConfigurationCallback {
...
+    this.prevData = {};
+    this.data["is-new-node"] = false;
+    this.data["is-removed"] = true;
+    this.cleanMetadata();
+    this.createPlaceholders();
+
+    // turn on Apply changes button as we need to indicate that section removal needs to be saved
+    this.iterateFields((att: string) : void => {
+      makeFieldDirty(this.meta[att]);
+    });
```
ryanemerson: Can be simplified to: 

```typescript
this.iterateFields((att) => makeFieldDirty(this.meta[att]));
```

vblagoje:
Awesome, thanks @ryanemerson. I'll prepare this commit without adding the ability to actually remove any sections. Then in a separate PR I'll add remove to specific sections. Some sections (not supported by ConfigurationSection component) will need custom work as well.

ryanemerson:
@vblagoje If you don't enable the removal of any sections, then how am I supposed to test if it works? ;)

Alternatively, why don't you just update this PR so that it applies to all common sections (indexing, expiration etc) that don't require any additional logic. You can then issue separate commits/PR for caches stores/loaders/security

vblagoje:
@ryanemerson done that. Please have a look and verify thoroughly. I will as well. Let's not introduce any new bugs.

vblagoje:
@ryanemerson If you create a new section, then delete it and finally try to save it - it fails.

vblagoje:
@ryanemerson should be ready now! Try to break it any way you can please! :-)

ryanemerson:
@vblagoje For Indexing we still have the option "NONE" in the select box, we should remove this as it's redundant. 

The undo operations no longer worked as expected. If a user enables expiration (only an example, this applies other sections as well) then all of the fields in that section are highlighted as modified and the undo link appears. However, if a user clicks undo nothing happens other than the css is reset. Maybe we should add an `undo (restart needed)` to the section title as well? and not show the undo option on individual fields until the user has entered a value. Not sure the best approach here tbh.

Related to ^, instead of copying the default values into the fields maybe we should utilise placeholder values? This might prevent the undo fields being triggered as well

vblagoje:
@ryanemerson yes we have NONE not only for indexing but for memory and load/store elements. Let's deal with those in a separate PR. Right after this one. How about this proposal now? I still initialize defaults, but without making fields dirty. Undo works. I added another switch to turn on "Apply changes" when appropriate! Have a look. If you agree let's adapt Indexing, memory and load/store in a separate commit and then we can integrate it all! Let me know wdyt

PR #186, opened 2017-02-06
ISPN-7443 Memory usage chart is not displayed

vblagoje:
Master only. @ryanemerson this is a regression due to id generation that I overlooked. We have to hard code this chart id and can not generate it on the fly.

### DIFF
--- a/src/module/server-instance/view/server-instance.html
+++ b/src/module/server-instance/view/server-instance.html
@@ -109,7 +109,7 @@
             <div class="row">
               <div class="col-md-8" ng-show="ctrl.serverInstance.isRunning()">
                 <h4>Heap Memory Usage</h4>
-                <div id-generator="perf_metrics.chart"></div>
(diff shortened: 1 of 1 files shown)


ryanemerson:
@vblagoje Can we use [c3's "bindTo"](http://c3js.org/gettingstarted.html#generate) config parameter so that we bind to a "chart" (or any name) class instead of an id? That way we can keep the unique id.

vblagoje:
@ryanemerson I tried that (using class="memory-chart") and then adjusting line 66 of ServerInstanceCtrl.ts . Please try it yourself as well.

ryanemerson:
@vblagoje I think I have it working, see https://github.com/infinispan/infinispan-management-console/pull/188

vblagoje:
Integrate with https://github.com/infinispan/infinispan-management-console/pull/188

PR #190, opened 2017-02-08
ISPN-7425 Enable admins to delete configuration subsections

vblagoje:
Master only. See https://issues.jboss.org/browse/ISPN-7425 for details

### DIFF
--- a/assets/languages/locale-enUS.txt
+++ b/assets/languages/locale-enUS.txt
@@ -160,5 +160,6 @@
   "YES":"Yes",
   "NEXT":"Next",
   "CLOSE":"Close",
-  "LOADING": "Loading"
+  "LOADING": "Loading",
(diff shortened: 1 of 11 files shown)


vblagoje:
@ryanemerson mind adjusting store and load :-) If you are super busy let me know and I'll do it! What about indexing. I think we can leave NONE as it is an actual option in metadata isn't it?

ryanemerson:
@vblagoje I agree about indexing. Regarding stores/loader, I might get a chance next week, but I'm focusing on some server changes this week.

Integrated with f6f97e2745800299be8a6059769da9af6b343dab, thanks.

vblagoje:
@ryanemerson ok no rush. I'll see if I can do it this week perhaps. We need to track these changes for a potential move downstream

PR #191, opened 2017-02-13
Change some IDs to static rather than generated - for better testability

rmacor:
Change some of the generated IDs to static IDs - for better testability


ryanemerson:
@rmacor thanks :)

PR #192, opened 2017-02-15
ISPN-7477 Endpoints with non-default names are not parsed

vblagoje:
Master only. See https://issues.jboss.org/browse/ISPN-7477 for details, rationale and test case.

### DIFF
--- a/src/common/utils/Utils.ts
+++ b/src/common/utils/Utils.ts
@@ -113,7 +113,7 @@ export function deepValue(object: Object, path: any): any {
   if (isArray(path)) {
     pathArray = path;
   } else if (isString(path)) {
-    pathArray = path.split(".")
+    pathArray = path.split(".");
   }
   for (var i: number = 0, len: number = pathArray.length; i < len; i++) {
     if (isNotNullOrUndefined(object)) {
--- a/src/common/utils/Utils.ts
+++ b/src/common/utils/Utils.ts
@@ -51,6 +51,17 @@ export function getArraySize(array: string[]): number {
   return 0;
 }
 
+export function traverseObjectRecursively(obj: any, callback: (visitedObject: any) => void): void {
+  if (isNotNullOrUndefined(obj)) {
+    for (var k in obj) {
+      if (isNotNullOrUndefined(obj[k]) && isObject(obj[k])) {
+        callback(obj[k]);
+        traverseObjectRecursively(obj[k], callback);
+      }
+    }
+  }
+}
+
 export function isArray(value: any): boolean {
   return Object.prototype.toString.call(value) === "[object Array]";
 }
@@ -97,8 +108,14 @@ export function deepGet(object: Object, path: string): any {
   return o;
 }
 
-export function deepValue(object: Object, path: string): any {
-  for (var i: number = 0, pathArray: string[] = path.split("."), len: number = pathArray.length; i < len; i++) {
+export function deepValue(object: Object, path: any): any {
+  let pathArray: string [] = [];
+  if (isArray(path)) {
+    pathArray = path;
+  } else if (isString(path)) {
+    pathArray = path.split(".")
+  }
+  for (var i: number = 0, len: number = pathArray.length; i < len; i++) {
     if (isNotNullOrUndefined(object)) {
       object = object[pathArray[i]];
     } else {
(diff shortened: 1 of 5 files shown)


src/services/endpoint/EndpointService.ts
```diff
@@ -27,7 +28,7 @@ export class EndpointService {
               private socketBindingService: SocketBindingService, private launchType: LaunchTypeService) {
   }
 
-  getAllEndpoints(profileName: string, socketBindingGroup: string): ng.IPromise<IEndpoint[]> {
+  getAllEndpoints(profileName: string, socketBindingGroup: string, cacheContainer:ICacheContainer): ng.IPromise<IEndpoint[]> {
```
ryanemerson: There is no need to pass all three of these arguments as they are all contained within ICacheContainer.

src/common/utils/Utils.ts
```diff
@@ -51,6 +51,17 @@ export function getArraySize(array: string[]): number {
   return 0;
 }
 
+export function traverseObjectRecursively(obj: any, callback: (visitedObject: any) => void): void {
```
ryanemerson: would a more meaningful name be something along the lines of "invokeCallbackRecursively"?

Also, we already have the `traverse` method in utils, would it be possible to utilise that?

src/services/endpoint/EndpointService.ts
```diff
@@ -37,12 +38,19 @@ export class EndpointService {
...
+        let isEndpoint: boolean = isNotNullOrUndefined(visitedObject["cache-container"]);
+        let isMultiRouterEndpoint: boolean = isNotNullOrUndefined(visitedObject["rest-socket-binding"])
+          || isNotNullOrUndefined(visitedObject["hotrod-socket-binding"]);
+        if (isEndpoint) {
+          let endpoint: IEndpoint = EndpointService.parseEndpoint(visitedObject.name, visitedObject);
+          if (endpoint["cache-container"] === cacheContainer.name) {
+            endpoints.push(endpoint);
+          }
+        } else if (isMultiRouterEndpoint) {
+          //todo parse multi-tenant router
```
ryanemerson: Is there anything stopping us catering for mult-tenant routers now? If not, I think it makes sense for us to get this right before we integrate.

src/services/endpoint/EndpointService.ts
vblagoje: +1 spot on!

src/common/utils/Utils.ts
vblagoje: @ryanemerson It would, but how do I access object enclosing key/value pairs? I could not figure it out

src/services/endpoint/EndpointService.ts
vblagoje: More work...let me see how much @ryanemerson

src/common/utils/Utils.ts
ryanemerson: @vblagoje See [ErrorModalCtrl](https://github.com/infinispan/infinispan-management-console/blob/master/src/common/dialogs/ErrorModalCtrl.ts#L15-L18) for an example. The parameter `value` in your callback is then equivalent to `obj[k]` in your function. 

Furthermore, we could improve the `traverse` function so that the callback parameter is more rigidly controlled like in your implementation. So instead of `callback: Function` we have:

```Typescript
export function traverse(obj: any, callback: (obj: any, value: any, trail: any[]) => void, trail?: any[]): void {
// content
}
```

vblagoje:
@ryanemerson I adapted the approach to use your recursion method. I also made a few fixes and cleanup a bit. I tested this new approach by making an actual multitenant router.
screen shot 2017-03-02 at 9 23 12 am

vblagoje:
@ryanemerson the question for you is: how to display multitenant router on cache containers page?

ryanemerson:
@vblagoje I think we should display the entry points for the router, something like

```
multi-tenancy router
    <hotrod-socket-binding-name>:<port>
    <rest-socket-binding-name>:<port>
```

Or even just: `multi-tenancy-router`

vblagoje:
@ryanemerson I added multi-tenancy router and I tried to display it as you suggested (and what I agree with) but the final string always gets displayed in one line no matter what I try. Can you please see how we can display it in multiple lines? If not we can make a new one line representation.

ryanemerson:
@vblagoje I believe the text is always being shown over one line because you're not returning html, so instead of `\n` you will need `<br />`. 

Another option is to have a method `isMultiRouter(endpoint)` and if true display the html for a multi-router otherwise display the normal endpoint html. This can be achieved by using `ng-if` on the divs containing the endpoint/multi-router html.

vblagoje:
@ryanemerson but that approach displays only the actual <br/>. I'll try the isMultiRouter approach.

vblagoje:
@ryanemerson here it is as we agreed. Please check that regular non-router endpoint list still work ok
screen shot 2017-03-09 at 8 07 12 am

ryanemerson:
@vblagoje Integrated https://github.com/infinispan/infinispan-management-console/commit/aec24b380925a0c43f94ec0ed4fbe4bbdc0d61f4, thanks!

PR #193, opened 2017-02-16
ISPN-7478: Enable the console to utilise Infinispan extension on Wild…

ryanemerson:
…fly.

https://issues.jboss.org/browse/ISPN-7478

@vblagoje These changes are required so that we account for profiles that may not have the infinispan subsystem defined. This can happen when a user installs the infinispan extension onto a standalone wildfly instance. For testing, I don't think it is necessary for you to go through the hassle of deploying the infinispan extension on wildfly as I have already checked this. However, it is important that you ensure these changes don't break the console in normal operation.


vblagoje:
Integrated with f3e997c Thanks @ryanemerson

PR #194, opened 2017-02-23
ISPN-7462 Server in reload-required state is not working correctly

vblagoje:
Master only. See https://issues.jboss.org/browse/ISPN-7462 for more details and how to force server into reload required state. This is a proposal, we can fine tune menu items if needed. Let's talk @ryanemerson


src/module/server-instance/view/server-instance.html
```diff
@@ -17,9 +17,9 @@
                 uib-dropdown-toggle><span translate>ACTIONS</span><span class="caret"></span>
         </button>
         <ul ng-if="ctrl.isDomainMode()" class="dropdown-menu" role="menu" aria-labelledby="dropdownMenu">
-          <li role="presentation" ng-class="ctrl.serverInstance.isRunning()?'disabled':''"><a id-generator="dropdown_menu.start" role="menuitem" ng-click="ctrl.createStartModal()" translate>START</a></li>
-          <li role="presentation" ng-class="!ctrl.serverInstance.isRunning()?'disabled':''"><a id-generator="dropdown_menu.stop" role="menuitem" ng-click="ctrl.createStopModal()" translate>STOP</a></li>
-          <li role="presentation" ng-class="ctrl.serverInstance.isRunning()?'disabled':''"><a id-generator="dropdown_menu.remove" role="menuitem" ng-click="ctrl.createRemoveModal()" translate>REMOVE</a></li>
+          <li role="presentation" ng-class="ctrl.canStart(ctrl.serverInstance)?'':'disabled'"><a id-generator="dropdown_menu.start" role="menuitem" ng-click="ctrl.createStartModal(ctrl.serverInstance)" translate>START_RELOAD_RESTART</a></li>
```
ryanemerson: Are Reload, Restart and Start mutually exclusive? Or is a state possible where both Reload and Restart are valid?

At the minute, if a user clicks the "Start/Reload/Restart" option they don't know what's actually going to happen behind the scenes. If the actions are mutually exclusive then we should just display the appropriate label (we could re-use `getActionRequired` for this. Otherwise we should display all options as separate menu items and only enable the valid options.

vblagoje: They are mutually exclusive. At least from what I have seen and experimented. If the state is reload-required then only reload is a valid action. If restart-required then only restart is a valid action. Same for stopped state. On start can start the server from that state. Not reload nor restart!

vblagoje: Ok then I'll make the menu item dynamic. I thought that was a bit weird. That's why I made this one :-) But I agree we can make it dynamic!

ryanemerson: Yeah maybe being completely dyanamic would be strange. Three separate menu items that are always present and then only enabled when appropriate could be less confusing? That way the user knows of those options, but also knows that only the enabled menu entries are appropriate for the current state

vblagoje: What do you think @pzapataf @tristantarrant @rmacor ?

rmacor: I prefer three separate buttons and only the one that is valid  would be enabled. One dynamic button would also work in my opinion, but static text - Start/Reload/Restart seems confusing. (the user wouldn't know which action is going to be performed).

As @ryanemerson mentioned I think it would be best if this state would also be propagated to cluster status and to server status indications on cluster page - there shouldn't be green check icon if the status is "reload-required"

I think "stop" action should be enabled when the server status is "reload-required". (the "stop" button is enabled in cluster actions)

vblagoje: Thanks @rmacor Let's go with three buttons then. Let's make another JIRA for properly indicating the state on server cards. I agree with your "stop" suggestion as well.

rmacor: Alright, I've created new JIRA about server cards https://issues.jboss.org/browse/ISPN-7544

vblagoje:
@ryanemerson we'll do cluster status in a separate JIRA https://issues.jboss.org/browse/ISPN-7544 Here we are dealing with three dynamically enabled menu items. You can enter reload-required state by following instructions at https://issues.jboss.org/browse/ISPN-7462

src/module/server-instance/view/server-instance.html
vblagoje: Thanks @rmacor

src/services/server/Server.ts
```diff
@@ -17,6 +17,13 @@ export class Server implements IServer {
     return this.state.toUpperCase() === "RUNNING";
   }
 
+  isReloadRequired(): boolean {
+    return this.state.toUpperCase() == "RELOAD-REQUIRED";
+  }
+  isRestartRequired(): boolean {
+    return this.state.toUpperCase() == "RESTART-REQUIRED";
```
ryanemerson: `isReloadRequired` and `isRestartRequired` should use `===`

src/module/server-instance/ServerInstanceCtrl.ts
```diff
@@ -113,6 +116,23 @@ export class ServerInstanceCtrl {
     return this.launchType.isDomainMode();
   }
 
+  private capitalizeFirstLetter(s: string): string {
+    return s.charAt(0).toUpperCase() + s.slice(1);
```
ryanemerson: Is it worth adding this method to Utils.ts? I can imagine us requiring it again

vblagoje: +1 yap @ryanemerson

src/services/server/Server.ts
vblagoje: +1

vblagoje:
@ryanemerson thanks for the insights. It should be gtg now

ryanemerson:
@vblagoje  Integrated https://github.com/infinispan/infinispan-management-console/commit/fea8c6b472f3e081e5f1d2d9126118f638b0ebbf, thanks!

PR #195, opened 2017-02-27
ISPN-7228 Creating wrong schema doesn't show error message (addendum)

vblagoje:
Master only. Patch for reopened ISPN-7228. Please review @ryanemerson

### DIFF
--- a/src/services/schemas/SchemasService.ts
+++ b/src/services/schemas/SchemasService.ts
@@ -67,6 +67,7 @@ export class SchemaService {
         error => deferred.reject(error))
       .then(error => {
           if (isNotNullOrUndefined(error)) {
+            this.unregisterProtoSchema(container, schema.fileName);
             deferred.reject(error);
           } else {
             deferred.resolve();


src/services/schemas/SchemasService.ts
```diff
@@ -67,6 +67,7 @@ export class SchemaService {
         error => deferred.reject(error))
       .then(error => {
           if (isNotNullOrUndefined(error)) {
+            this.unregisterProtoSchema(container, schema.fileName);
```
ryanemerson: How important is this action? Do we need to take action if it fails?

vblagoje: We should take it only if we get an error @ryanemerson See reopened ISPN-7228 for details

ryanemerson: So at present this is an async action that we fire and forget, no action is taken if the unregister fails/timesout. Do we want to retry if it fails? Or at least notify the user that the schema was created, but with errors and it hasn't been possible to undo its creation

vblagoje: I don't think we need to retry. How many times should we retry? The failure of not being able to remove schema is not a big deal @ryanemerson

vblagoje: @ryanemerson he/she can always remove it!

ryanemerson: @vblagoje Ok cool, integrating

PR #199, opened 2017-03-19
ISPN-7628 The cluster status doesn't reflect reload-required state of its nodes

vblagoje:
Master only. Please review and integrate @ryanemerson

### DIFF
--- a/src/module/server-group/ServerGroupCtrl.ts
+++ b/src/module/server-group/ServerGroupCtrl.ts
@@ -20,7 +20,7 @@ import {
 
 export class ServerGroupCtrl {
   static $inject: string[] = ["$state", "$uibModal", "dmrService", "serverGroupService", "serverService",
-    "jGroupsService", "launchType", "serverGroup", "available", "runningInstances"];
+    "jGroupsService", "launchType", "serverGroup", "available", "runningInstances", "status"];
 
   status: string = "DEGRADED";
   serverStatusMap: IMap<string> = {};
@@ -37,12 +37,13 @@ export class ServerGroupCtrl {
               private launchType: LaunchTypeService,
               public serverGroup: IServerGroup,
               public available: boolean,
-              public runningInstances:IServerAddress[]) {
+              public runningInstances:IServerAddress[],
+              private status: IMap<string>) {
     this.fetchSGCoordinator();
     this.fetchServerStatuses();
     this.fetchInetAddresses();
     this.hosts = this.filterUniqueHosts();
-    this.status = available ? "STARTED" : (isNotNullOrUndefined(runningInstances) && runningInstances.length > 0) ? "DEGRADED" : "STOPPED";
+    this.status = status;
   }
 
   isCoordinator(server: IServerAddress): boolean {
@@ -51,6 +52,32 @@ export class ServerGroupCtrl {
 
   isServerRunning(server: IServerAddress): boolean {
     return this.isServerInState(server, SERVER_STATE_RUNNING);
+  }
+
+  areAllServersInRunningState(): boolean {
+    let result:boolean = this.areAllServersInServerGroupInState(SERVER_STATE_RUNNING);
+    return result;
+  }
+
+  areAllServersInReloadRequiredState(): boolean {
+    let result:boolean = this.areAllServersInServerGroupInState(SERVER_STATE_RELOAD_REQUIRED);
+    return result;
+  }
+
+  areAllServersInRestartRequiredState(): boolean {
+    let result:boolean = this.areAllServersInServerGroupInState(SERVER_STATE_RESTART_REQUIRED);
+    return result;
+  }
+
+  areAllServersInStoppedOrOtherState(): boolean {
+    return !this.areAllServersInRunningState() &&
+      !this.areAllServersInReloadRequiredState() &&
+        !this.areAllServersInRestartRequiredState();
+  }
+
+  areAllServersInStoppedState(): boolean {
+    let result:boolean = this.areAllServersInServerGroupInState(SERVER_STATE_STOPPED);
+    return result;
   }
 
   isServerStopped(server: IServerAddress): boolean {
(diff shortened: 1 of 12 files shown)


vblagoje:
screen shot 2017-03-19 at 7 19 20 am
screen shot 2017-03-19 at 7 20 31 am
screen shot 2017-03-19 at 7 20 40 am

vblagoje:
@rmacor @ryanemerson I have adjusted all cluster states and added appropriate labels, dialogs, menu items etc. I have also changed state from started to running to closer reflect the cluster state. See https://issues.jboss.org/browse/ISPN-7628 for more details and comments. I am open to suggestions, let's solve this one appropriately.

vblagoje:
I updated PR @ryanemerson and @rmacor . Have a look when you get a chance

src/module/server-group/ServerGroupCtrl.ts
```diff
@@ -53,6 +53,31 @@ export class ServerGroupCtrl {
...
+
+  isAtLeastOneServerInStoppedState(): boolean {
+    let result: boolean = this.isAtLeastOneServerInServerGroupInState(SERVER_STATE_STOPPED);
+    return result;
+  }
+
+  areAllServersInStoppedState(): boolean {
+    let result: boolean = this.areAllServersInServerGroupInState(SERVER_STATE_STOPPED);
+    return result;
+  }
```
ryanemerson: There's no need for the `result` variable in any of the above methods.

src/module/server-group/ServerGroupCtrl.ts
```diff
@@ -174,6 +215,26 @@ export class ServerGroupCtrl {
     return this.launchType.isDomainMode();
   }
 
+  private areAllServersInServerGroupInState(state: string): boolean {
+    let serverStatuses: string [] = this.serverStatuses();
```
ryanemerson: No need for local variable

src/module/server-group/ServerGroupCtrl.ts
```diff
@@ -174,6 +215,26 @@ export class ServerGroupCtrl {
...
+    return serverStatuses.every((serverStatus: string) => {
+      return serverStatus === state;
+    });
+  }
+
+  private isAtLeastOneServerInServerGroupInState(state: string): boolean {
+    let serverStatuses: string [] = this.serverStatuses();
+    return serverStatuses.indexOf(state) > -1;
```
ryanemerson: We can just call `this.serverStatuses()` direct

src/services/server-group/ServerGroupService.ts
```diff
@@ -233,6 +236,26 @@ export class ServerGroupService {
     return this.getStringFromAllMembers(serverGroup, (server) => this.serverService.getServerStatus(server));
   }
 
+  getServerGroupStatus(serverGroup: IServerGroup): ng.IPromise<string> {
+    return this.getServerStatuses(serverGroup).then((statuses) => {
+      let status: string = "DEGRADED";
+      let statusArray: string[] = [];
+      for (let serverKey in statuses) {
+        statusArray.push(statuses[serverKey].toUpperCase());
+      }
```
ryanemerson: The logic to generate a status array is now duplicated here and `ServerGroupCtrl`. We should expose this as a single method in `ServerGroupService`

src/services/server-group/ServerGroupService.ts
```diff
@@ -233,6 +236,26 @@ export class ServerGroupService {
...
+      let needsRestart = statusArray.indexOf(SERVER_STATE_RESTART_REQUIRED) > -1;
+      let needsReload = statusArray.indexOf(SERVER_STATE_RELOAD_REQUIRED) > -1;
```
ryanemerson: Missing type definitions on the two variables

src/module/server-group/ServerGroupCtrl.ts
vblagoje: +1 I used it for easier debugging. Now, I'll remove it!

vblagoje:
Updated, thanks for the feedback @ryanemerson I could not reproduce stopping the cluster error.

rmacor:
One more suggestion, should start action be available if there is at least one node in stopped state? Currently it's only available if all the nodes are stopped.

vblagoje:
What we have now @rmacor is that we keep cluster in "running" state if there is at least one server running. Say you have A and B. If B is shut down, a cluster is still running. So having a start action is confusing. At least according to me :-) We now have the opposite logic. If there is at least one server running we allow the stop action. If all servers are stopped we allow start. Thoughts @ryanemerson ?

ryanemerson:
@vblagoje @rmacor I think the current logic makes sense with the start/stop labels, as you the user can still click on an individual server state to start that single instance.

src/module/server-group/ServerGroupCtrl.ts
```diff
@@ -53,6 +53,26 @@ export class ServerGroupCtrl {
...
+  isAtLeastOneServerInRestartRequiredState(): boolean {
+    return this.isAtLeastOneServerInServerGroupInState(SERVER_STATE_RESTART_REQUIRED);
+  }
+
+  isAtLeastOneServerInStoppedState(): boolean {
+    return this.isAtLeastOneServerInServerGroupInState(SERVER_STATE_STOPPED);
+  }
+
+  areAllServersInStoppedState(): boolean {
+    return this.areAllServersInServerGroupInState(SERVER_STATE_STOPPED);
```
ryanemerson: @vblagoje All of these methods will result in a call to the server to get the status map array, so each page load will make the same call at least 5 times. It would be better if we pulled the status map once upon page load. When a user clicks stop/start/etc for a cluster, we should then force a page refresh (I think we already do this in most cases) and the user will be presented with the updated status

vblagoje: +1 ok @ryanemerson makes sense

src/module/server-group/ServerGroupCtrl.ts
```diff
@@ -37,12 +37,12 @@ export class ServerGroupCtrl {
               private launchType: LaunchTypeService,
               public serverGroup: IServerGroup,
               public available: boolean,
-              public runningInstances:IServerAddress[]) {
+              public runningInstances:IServerAddress[],
```
ryanemerson: AFAICT `runningInstances` is no longer utilised anywhere

src/module/server-group/ServerGroupCtrl.ts
vblagoje: But wait @ryanemerson that's not the case! ``statusMapToArray`` method just takes the IMap and converts it to the array. There is no server call here, right?

ryanemerson: O yeah, my mistake. In that case we should only call `statusMapToArray` once in `fetchServerStatuses` or controller's constructor and store is a class variable.

vblagoje: Ok, done. Have another thorough look @ryanemerson but let's wait for integration until we get consensus about these states! I want to get this one in - once without subsequent additions.

ryanemerson:
@vblagoje Integrated with 955e3f78b74ea0f7b78d46083c97ee522d869e30, thanks!

PR #204, opened 2017-04-13
ISPN-7705 Greyed out cluster action can still be selected

vblagoje:
Master only. See https://issues.jboss.org/browse/ISPN-7705 for more details.


vblagoje:
@ryanemerson also notice how we use id instead of id-generator?

ryanemerson:
@vblagoje We use Id for the menu items as that was easier for Roman, it was change as part of https://github.com/infinispan/infinispan-management-console/commit/55bb3f67b8ca559ac0b1c475d77398c005e27471

PR #209, opened 2017-06-07
Endpoints - first version

vblagoje:
Master only. We can migrate downstream once we have full endpoints support with ISPN-7881 included.


src/services/dmr/DmrService.ts
```diff
@@ -171,6 +190,101 @@ export class DmrService {
...
+    }
+  }
+
+  private parseJson(value: string, key: string): string {
+    try {
+      return JSON.parse(value);
+    } catch (e) {
+      console.log("Invalid JSON value " + value + " for field " + key);
+    }
+  }
```
ryanemerson: This appears to be identical to the code found in `CacheConfigService`. We should create a suitable abstraction so that this code is not repeated unnecessarily. For example, we could create an `AbstractConfigService` which contains this logic and then make `EndpointService` and `CacheConfigService` extend it.

src/module/server-group/endpoints/config/EndpointConfigCtrl.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  create(endpoint: IEndpoint): ng.IPromise<any> {
+    let excludedAttributes: string [] = ["is-new-node"];
+    return this.endpointService.save(endpoint, excludedAttributes);
+  }
+
+  update(endpoint: IEndpoint): ng.IPromise<any> {
+    let excludedAttributes: string [] = ["socket-binding", "is-new-node"];
+    return this.endpointService.save(endpoint, excludedAttributes);
+  }
```
vblagoje: Yes, but we need different exclude exclusion lists for save and update. I thought of this way as the best approach to differentiate the two. EndpointsService should not care about specific exclusion lists, higher API that calls it - should. Does that make more sense?

src/module/server-group/endpoints/config/EndpointConfigCtrl.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  isEditMode(): boolean {
+    return this.$state.current.name === "edit-endpoint-config";
+  }
+
+  isTemplateNameEmpty(): boolean {
+    //let templateName: string = this.template["template-name"];
+    //return !(isNotNullOrUndefined(templateName) && isNonEmptyString(templateName));
+    return true;
+  }
```
vblagoje: +1

src/module/cache-containers/view/cache-containers.html
```diff
@@ -28,8 +28,10 @@ <h1 translate>CACHE_CONTAINERS</h1>
             <div id-generator="{{ctrl.getContainerId('endpoints', container)}}" class="endpoint" ng-repeat="endpoint in container.endpoints">
                 <div ng-if="ctrl.isMultiTenantRouter(endpoint)">
                   <span>multi-tenancy router</span></br>
+                    <!--
                     <span ng-if="endpoint['hotrod-socket-binding']">&nbsp;&nbsp;{{endpoint['hotrod-socket-binding'].name + " : " + endpoint['hotrod-socket-binding'].port}}</span></br>
                     <span ng-if="endpoint['rest-socket-binding']">&nbsp;&nbsp;{{endpoint['rest-socket-binding'].name + " : " + endpoint['rest-socket-binding'].port}}</span></br>
+                    -->
```
vblagoje: Endpoint object changed so I just commented out the old code, I'll update it!

src/module/server-group/endpoints/AddEndpointModalCtrl.ts
```diff
@@ -0,0 +1,23 @@
+
+export class AddEndpointModalCtrl{
+  static $inject: string[] = ["serverGroup", "$scope", "$state"];
+
+  public endpointTypes: string[] = ["hotrod", "rest", "websocket", "router", "memcached"];
```
vblagoje: Good idea

src/module/server-group/endpoints/EndpointsCtrl.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export class EndpointsCtrl {
+  static $inject: string[] = ["$uibModal", "$state", "endpointService", "dmrService", "serverGroup", "endpoints"];
+
+  constructor(private $uibModal: IModalService,
+              private $state: IStateService,
+              private endpointService: EndpointService,
+              private dmrService: DmrService,
+              private serverGroup: IServerGroup,
+              private endpoints: IEndpoint[]){
```
vblagoje: Right on

src/module/server-group/endpoints/config/EndpointConfigCtrl.ts
ryanemerson: I understand the rationale, but where else are these methods required? Isn't it always the same things that should be excluded when saving/updating an endpoint? 

Also, with my suggestion can you still specify a different exclusion list by calling `EndpointService.save(endpoint, ["example"]);`

vblagoje:
@ryanemerson I'll hide name field from the UI as we can use the name of the actual endpoint as it's unique name. If you look into DMR name attribute isn't even mandatory. In a demo yesterday with @pzapataf and @tristantarrant we agreed to remove the name. It is just causing massive confusion

vblagoje:
@ryan "It is possible to click the "Create" button for a connector even if the Cache Container is empty, this results in an error being thrown by the server" - we don't have a method to mark a field as mandatory, as of today? If not, we can tackle this later.

vblagoje:
"On the ..cluster/endpoints summary page, if a name is configured for the socket, it currently isn't displayed on this summary page. Should it be so that we can better distinguish between multiple connectors of the same type?" I'll what I can do here.

src/module/server-group/endpoints/config/EndpointConfigCtrl.ts
vblagoje: Gotcha. I'll make the change.

vblagoje:
@ryanemerson I added the two commits that migrate console to Patternfly 4.0.0.rc1 as well. Then we can append the single UI/X commit from @nmedarevic

vblagoje:
@ryanemerson I also want to split the PR review into commits that are relevant to the specific commit rather than have a lump PR review commit on top. I'll do that part myself.

src/services/dmr/DmrService.ts
vblagoje: @ryanemerson what I meant to do is to retire all the mumbo jumbo DMR code from CacheConfigService and use the new model saving from DMRService. I'll do that later after this initial transition as a part of regular refactoring.

vblagoje:
@ryanemerson @tristantarrant @pzapataf I discovered that name attribute of an endpoint is not irrelevant as we thought. Configuration file standalone.xml/domain.xml does not name a specific connector by the given DMR name. If we have two connectors of the same type and we do not give them a name then when we restart the server - if fails. For the endpoint creation use case, I'll automatically assign DMR path name to endpoint name attribute.

vblagoje:
@ryanemerson I added handling of a use case when we need to create an endpoint from ui and then copy its name into name attribute. That created a problem with an existing check from configuration section constructor - i.e hasAnyFieldPreviousData invoked in constructor that determines if the node is new. If you see a better way to handle this - I would love to hear.

vblagoje:
@ryanemerson I also could not make endpoint "name" field UI read-only - anyway I tried. I need your help there.

ryanemerson:
@vblagoje For making fields read-only, see the [CacheTemplatesCtrl](https://github.com/infinispan/infinispan-management-console/blob/master/src/module/cache-templates/CacheTemplatesCtrl.ts#L33) as an example. In this case we specify the array of field names that are to be readOnly in the controller and then pass them to the cache-configuration component via the [view](https://github.com/infinispan/infinispan-management-console/blob/master/src/module/cache-templates/view/cache-templates.html#L35)

src/module/server-group/ServerGroup.ts
```diff
@@ -34,4 +39,65 @@ module.config(($stateProvider: ng.ui.IStateProvider) => {
...
+    controller: EndpointConfigCtrl,
+    controllerAs: "ctrl",
+    templateUrl: "module/server-group/endpoints/config/view/endpoint-config.html",
+    resolve: {
+      serverGroup: ["$stateParams", "serverGroupService", ($stateParams, serverGroupService) => {
+        return serverGroupService.getServerGroupMapWithMembers($stateParams.serverGroup);
+      }],
+      endpoint: ["$stateParams", ($stateParams) => {
+        let newEndpoint:IEndpoint = EndpointService.parseEndpoint([].concat($stateParams.endpointType).concat($stateParams.endpointName), {});
+        newEndpoint.createdFromUI();
```
ryanemerson: This is redundant as you already know whether you're editing the endpoint or creating a new one via the $state in the `EndpointConfigCtrl`. So you could just do `!isEditMode()` or be more explicit and create a `isCreateMode()` which checks the $state. It's better to keep as much of this logic in the Controller as possible as resolve should just be for fetching data required before page load.

src/module/server-group/ServerGroup.ts
```diff
@@ -34,4 +39,65 @@ module.config(($stateProvider: ng.ui.IStateProvider) => {
...
+      endpoints: ["endpointService", "serverGroup",
+        (endpointService: EndpointService, serverGroup: IServerGroup) => {
+          return endpointService.getAllClusterEndpoints(serverGroup);
+        }],
+    }
+  });
+
+  $stateProvider.state("edit-endpoint-config", {
+    parent: "root",
+    url: "/server-groups/:serverGroup/endpoints/:endpointType/:endpointName/config",
```
ryanemerson: This should be different once we have exposed an Endpoint module, but when inheriting from root you should start with "/" otherwise we get "http://localhost:3000/#//server-groups/cluster/endpoints/hotrod-connector/hotrod-connector/config"

src/module/server-group/ServerGroup.ts
vblagoje: Yes, the problem is that I don't want any of this logic to leak into configuration section. I agree with your suggestion above!

vblagoje:
@ryanemerson figured out read-only fields.

vblagoje:
@ryanemerson as per our consultation please have a look at the last commit with the changes you suggested.

src/services/endpoint/EndpointService.ts
```diff
@@ -8,31 +8,31 @@ import IQService = angular.IQService;
...
+  static parseEndpoint(namePath: string [], object: any, socketBinding?: ISocketBinding): IEndpoint {
+    return new Endpoint(namePath, object, socketBinding);
   }
 
   constructor(private $q: IQService, private dmrService: DmrService,
               private socketBindingService: SocketBindingService, private launchType: LaunchTypeService) {
   }
 
+  createEndpoint(namePath: string []): IEndpoint {
+    return new Endpoint(namePath, {"is-new-node": true}, undefined);
```
ryanemerson: This method should be static. Also, if we simply pass undefined for the second arg, then we don't need [this](https://github.com/vblagoje/infinispan-management-console/blob/7f169f8473a89b9fed7067d013a1bf5e3eb3627a/src/components/configuration-section/ConfigurationSectionCtrl.ts#L34) hack. This may need slight changes elsewhere, but the `configuration-section` was designed to handle the creation of new configurations as we do with templates etc.

src/module/endpoints/Endpoints.ts
```diff
@@ -0,0 +1,51 @@
...
+import {EndpointConfigCtrl} from "./../endpoints/config/EndpointConfigCtrl";
+
+const module: ng.IModule = App.module("managementConsole.endpoints", []);
+
+
+module.config(($stateProvider: ng.ui.IStateProvider) => {
+
+  $stateProvider.state("edit-endpoint-config", {
+    parent: "root",
+    url: "endpoints/:profile/:endpointType/:endpointName/config",
```
vblagoje: Yes, but how do I do that? :-)

src/module/endpoints/Endpoints.ts
```diff
@@ -0,0 +1,51 @@
...
+        return serverGroupService.getServerGroupByProfile($stateParams.profile);
+      }],
+      endpointName: ["$stateParams", ($stateParams) => $stateParams.endpointName],
+      endpointType: ["$stateParams", ($stateParams) => $stateParams.endpointType]
+    }
+  });
+
+  $stateProvider.state("new-endpoint-config", {
+    parent: "root",
+    url: "endpoints/:profile/:endpointType/:endpointName/newEndpoint",
```
vblagoje: +1

src/module/endpoints/Endpoints.ts
ryanemerson: You need to create a route, say "endpoints" which implements IRedirectState and points to a child route in the `redirectTo` field. So in this case, "edit-endpoint-config" would become "endpoints.edit-endpoint.config" (you could just call it "endpoints.edit" for short). See the [CacheContainer](https://github.com/infinispan/infinispan-management-console/blob/master/src/module/cache-container/CacheContainer.ts#L23-L40) for an example.

src/services/endpoint/EndpointService.ts
vblagoje: How exactly? I would love to remove the hack but we need to somehow set data.name to the name of endpoint. However, hasAnyFieldPreviousData method in configuration section then marks the component with {"is-new-node": false} and then we cannot create Endpoint

vblagoje:
I don't see any new gulp check failures. How do we make tslint ignore "object access via string literals is disallowed"?

src/module/endpoints/Endpoints.ts
vblagoje: Ok, I get it.... but let's wait for this one as it is not essential now and we might still tune a bit this page. We can do the redirect at that time.

ryanemerson:
@vblagoje I would say we don't want to remove the "object via string literals" error as it is best practice to avoid string literals where possible as it makes the integration with the IDE better. The errors are currently in Endpoint.ts and these are easily changed.

src/services/endpoint/EndpointService.ts
ryanemerson: For Cache templates we make the `CacheTemplatesCtrl` call the update or create service depending on whether it is in edit or create mode. So in `EndpointConfigCtrl` you can just call `EndpointService::create` or `EndpointService::update` depending on what's appropriate. If you then need "is-new-node" setting for your recursive traversal, you can do that in ^ service, as AFAICT, you don't use "is-new-node" anywhere in the controller.

ryanemerson: Tbh though, we can stick with this for now as you're planning on making changes to how we interact with the DMR etc anyway.

vblagoje: Right @ryanemerson let me see what I can do here

vblagoje:
@ryanemerson ok, made the changes you requested. Have a look again

ryanemerson:
Integrated. Great work @vblagoje!

PR #211, opened 2017-06-14
ISPN-7881 Administration console - enhance field entry UI/X

vblagoje:
Preview, do not integrate yet @ryanemerson Waiting for other PRs to be merged and then to rebase.


src/components/forms/form-group/FormGroupCtrl.ts
```diff
@@ -24,17 +24,45 @@ export class FormGroupCtrl {
...
 
   constructor() {
+    this.multiSettings = {
+      template: "{{option}}",
+      smartButtonTextConverter(skip: any, option: any): any {
+        return option;
+      }
+    };
+    this.type = getTypeModelType(this.meta);
+    if (this.meta.type.TYPE_MODEL_VALUE === "LIST") {
```
vblagoje: @ryanemerson I found that I need to JSON.parse lists. Don't we do this JSON response parsing automatically?

ryanemerson: Hmm strange. I guess that when the JSON is parsed it is assuming that the list is a String, not an array, for some reason. What does the DMR output look like?

vblagoje: @ryanemerson Yes, exactly that. What do you mean DMR output? When I first loaded the model values into the field it sliced them into one item per character. So the dropdown menu had 20+ items. After I used JSON.parse it all started to work, both model loading and saving.

vblagoje:
@ryanemerson if you manage, have a look at this one as well :-)

ryanemerson:
@vblagoje I tried to resolve the conflicts locally to speed things up, but I kept running into the following issues when running `gulp serve`. I did run `jspm install`, but this did not seem to solve the problem.

```
src/ManagementConsole.ts(8,1): error TS6053: File '/home/remerson/workspace/RedHat/infinispan/infinispan-management-console/typings/modules/deepmerge/index.d.ts' not found.

src/services/metadata/MetadataService.ts(5,28): error TS2307: Cannot find module 'deepmerge'.

```

vblagoje:
No problem @ryanemerson let me look into it

vblagoje:
Have a look now @ryanemerson Test it on endpoints to see the new feature but also that it does not make a regression of form group (cache loaders/stores configuration)

ryanemerson:
@vblagoje I am coming across the following error when trying to set the "receiver-buffer-size" to 1 on the default hotrod-connector:

```json
{"WFLYCTL0062: Composite operation failed and was rolled back. Steps that failed:":{"Operation step-10":"WFLYCTL0175: Resource [\n (\"profile\" => \"clustered\"),\n (\"subsystem\" => \"datagrid-infinispan-endpoint\"),\n (\"hotrod-connector\" => \"hotrod-connector\"),\n (\"authentication\" => \"AUTHENTICATION\")\n] does not exist; a resource at address [\n (\"profile\" => \"clustered\"),\n (\"subsystem\" => \"datagrid-infinispan-endpoint\"),\n (\"hotrod-connector\" => \"hotrod-connector\"),\n (\"authentication\" => \"AUTHENTICATION\"),\n (\"sasl\" => \"SASL\")\n] cannot be created until all ancestor resources have been added"}}
```

vblagoje:
Should work now @ryanemerson but please check if there are any regressions cache configuration - where we deal with lists. I am referring to changes in FormGroupCtrl I made. If anything should break is cache loaders/stores.

ryanemerson:
@vblagoje The changes to FormGroup seem fine, the only problem I have noticed is when trying to remove a section, such as authentication->Sasl Configuration, via the garbage can button I get the following error:

```Typescript
angular.js:14202 TypeError: Cannot read property 'forward-secrecy' of undefined
    at ConfigurationSectionCtrl.eval (ConfigurationSectionCtrl.ts:43)
    at Array.some (<anonymous>)
    at eval (ConfigurationSectionCtrl.ts:43)
    at Array.some (<anonymous>)
    at ConfigurationSectionCtrl.hasAnyFieldPreviousData (ConfigurationSectionCtrl.ts:42)
    at fn (eval at compile (angular.js:NaN), <anonymous>:4:273)
    at Scope.$digest (angular.js:17831)
    at eval (angular.js:18036)
    at completeOutstandingRequest (angular.js:6048)
    at eval (angular.js:6327)
```

vblagoje:
Ok, let me see @ryanemerson

vblagoje:
@ryanemerson check it out now. These sections should not be removable, to begin with. Then I found a bug where removable is actually a string not a boolean. So the check had to be a bit different. Should be ok now.

ryanemerson:
Integrated 5c437df42644444d863d3199a4260c9eabc079d2, great work @vblagoje.

ryanemerson:
@vblagoje I've left the JIRA open as we will need to add the appropriate JSON etc to normal cache configuration pages eventually.

PR #212, opened 2017-06-30
ISPN-5812 Enable configuration of endpoints (addendum)

vblagoje:
@ryanemerson I hard coded the profile name on line 141 of EndpointService. We need profile encapsulated in Endpoint type. Please verify.


ryanemerson:
@vblagoje Good spot! Integrated, thanks

PR #214, opened 2017-07-04
ISPN-7276: Console should automatically logout on Server mode change

nmedarevic:
A proposal for an improvement based on the https://issues.jboss.org/browse/ISPN-7276 issue.

@ryanemerson @vblagoje Could you take a look?

### DIFF
--- a/src/services/authentication/AvailabilityCheck.ts
+++ b/src/services/authentication/AvailabilityCheck.ts
@@ -1,12 +1,21 @@
+import {App} from "../../ManagementConsole";
 import {DmrService} from "../dmr/DmrService";
 
+const module: ng.IModule = App.module("managementConsole.services.authentication");
+
 export class AvailabilityCheck {
+
+  static $inject: string[] = ["$interval", "dmrService", "$q", "$window", "$timeout"];
+  static currentServerLaunchType: string;
 
   private alive: boolean = true;
   private serverCheck: ng.IPromise<any>;
 
   constructor(private $interval: ng.IIntervalService,
-              private dmrService: DmrService) {
+              private dmrService: DmrService,
+              private $q: ng.IQService,
+              private $window: ng.IWindowService,
+              private $timeout: ng.ITimeoutService) {
   }
 
   startApiAccessibleCheck(): void {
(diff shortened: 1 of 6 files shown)


src/services/interceptors/ServerStatusInterceptor.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  public static serverStatusTimeout:number = 5000;
+  public static serverStatusTimer: any;
+  public static statusCallsFired: number = 0;
+  public static maxStatusCalls: number = 20;
+
+  public static ngInjector:any;
+  public static ngQ:any;
+
+  public static resetLastRequestTimer(timer: number, self: any): any {
```
ryanemerson: Is there any need for the methods and variables to be static?

src/services/authentication/AvailabilityCheck.ts
```diff
@@ -1,12 +1,21 @@
+import {App} from "../../ManagementConsole";
 import {DmrService} from "../dmr/DmrService";
 
+const module: ng.IModule = App.module("managementConsole.services.authentication");
+
 export class AvailabilityCheck {
 
+  static $inject: string[] = ["$interval", "dmrService", "$q", "$window", "$timeout"];
+  static currentServerLaunchType: string;
```
ryanemerson: Does this variable need to be static?

src/services/authentication/AvailabilityCheck.ts
```diff
@@ -21,8 +30,25 @@ export class AvailabilityCheck {
...
   }
 
-  private checkServerIsAlive(): void {
+  public checkServerIsAlive(): void {
     this.dmrService.readAttribute({address: [], name: "launch-type"})
-      .then(() => this.alive = true, () => this.alive = false);
+      .then((response) => {
+        if (!response) {
+          return undefined;
+        }
```
ryanemerson: Won't response always be a string? I'm not sure this check and return make sense

src/services/authentication/AvailabilityCheck.ts
```diff
@@ -21,8 +30,25 @@ export class AvailabilityCheck {
...
+
+        if (!AvailabilityCheck.currentServerLaunchType) {
```
ryanemerson: Maybe use `Utils::isNullOrUndefined` here

src/services/dmr/DmrService.ts
```diff
@@ -159,6 +159,7 @@ export class DmrService {
       .then(
         (success: any) => deferred.resolve(success.data.result),
         (failure: any) => {
+          console.log(failure);
```
ryanemerson: We used to log the failure output here, however we should be handling failures in a modal (or some other means). So I would rather not reintroduce this behaviour.

src/services/interceptors/ServerStatusInterceptor.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (self.isServerRequestStatusCall(config.config)) {
+      self.statusCallsFired = 0;
+      self.serverStatusTimer = self.resetLastRequestTimer(self.serverStatusTimer, self);
+    }
+
+    return config;
+  }
+
+  public response(config: any): any {
+    const self: any = ServerStatusInterceptor.self;
```
ryanemerson: You can just use the `this` keyword instead of defining a self const

src/services/interceptors/ServerStatusInterceptor.ts
```diff
@@ -0,0 +1,102 @@
...
+  public static ngInjector:any;
+  public static ngQ:any;
+
+  public static resetLastRequestTimer(timer: number, self: any): any {
+    if (timer) {
+      clearTimeout(timer);
+    }
+
+    return setTimeout(function(): void {
+      self.ngInjector.get("availabilityCheck").checkServerIsAlive();
```
ryanemerson: As you have now exposed AvailabilityCheck as a service, why can't you just inject the service via $inject?

ryanemerson:
I'm closing this PR as it is no longer active, thank you for your effort @nmedarevic. Contributions are always welcome :+1:

PR #217, opened 2017-07-17
ISPN-7649 Administration console - transaction tab allows to set invalid options

vblagoje:
Master only. See https://issues.jboss.org/browse/ISPN-7649

### DIFF
--- a/src/services/cache-config/CacheConfigService.ts
+++ b/src/services/cache-config/CacheConfigService.ts
@@ -303,34 +303,44 @@ export class CacheConfigService {
     let address: string[] = this.getConfigAddress(container.name, container.profile).concat(type + "-configuration", name);
     let builder: CompositeOpBuilder = new CompositeOpBuilder();
     let deferred: ng.IDeferred<void> = this.$q.defer<void>();
+    let validConfiguration: boolean = true;
+
+    try {
+      this.verifyConfiguration(config, deferred);
+    } catch (e) {
+      validConfiguration = false;
+      deferred.reject(e.message);
+    }
+
     config.template = template;
 
     let createNodeRequest: IDmrRequest = this.createAddOperation(address, config, ["name", "type", "template-name", "is-new-node"]);
 
-    this.dmrService.executePost(createNodeRequest)
-      .then(() => {
-        this.createHelper(builder, address.concat("locking", "LOCKING"), config.locking);
-        this.createHelper(builder, address.concat("memory", this.getMemoryType(config.memory)), config.memory);
-        this.createHelper(builder, address.concat("expiration", "EXPIRATION"), config.expiration);
-        this.createHelper(builder, address.concat("indexing", "INDEXING"), config.indexing);
-        this.createHelper(builder, address.concat("compatibility", "COMPATIBILITY"), config.compatibility);
-        this.createHelper(builder, address.concat("partition-handling", "PARTITION_HANDLING"), config["partition-handling"]);
-        this.createHelper(builder, address.concat("transaction", "TRANSACTION"), config.transaction);
-        this.createHelper(builder, address.concat("state-transfer", "STATE_TRANSFER"), config["state-transfer"]);
-        this.createBackupNodes(builder, address.concat("backup"), config.backup);
+    if (validConfiguration) {
+      this.dmrService.executePost(createNodeRequest)
+        .then(() => {
+          this.createHelper(builder, address.concat("locking", "LOCKING"), config.locking);
+          this.createHelper(builder, address.concat("memory", this.getMemoryType(config.memory)), config.memory);
+          this.createHelper(builder, address.concat("expiration", "EXPIRATION"), config.expiration);
+          this.createHelper(builder, address.concat("indexing", "INDEXING"), config.indexing);
+          this.createHelper(builder, address.concat("compatibility", "COMPATIBILITY"), config.compatibility);
+          this.createHelper(builder, address.concat("partition-handling", "PARTITION_HANDLING"), config["partition-handling"]);
+          this.createHelper(builder, address.concat("transaction", "TRANSACTION"), config.transaction);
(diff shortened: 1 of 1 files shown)


src/services/cache-config/CacheConfigService.ts
```diff
@@ -315,10 +315,13 @@ export class CacheConfigService {
         this.createHelper(builder, address.concat("indexing", "INDEXING"), config.indexing);
         this.createHelper(builder, address.concat("compatibility", "COMPATIBILITY"), config.compatibility);
         this.createHelper(builder, address.concat("partition-handling", "PARTITION_HANDLING"), config["partition-handling"]);
-        this.createHelper(builder, address.concat("transaction", "TRANSACTION"), config.transaction);
+        if (this.isValidTXMode(builder, address, config)) {
+          this.createHelper(builder, address.concat("transaction", "TRANSACTION"), config.transaction);
+        } else {
+          deferred.reject("Cannot set TX mode to FULL_XA with ASYNC cache mode");
+        }
```
ryanemerson: Should we place this logic in some kind of `createTransactions` method, similar to `createCacheStore` etc, so that the logic remains self contained?

Also, do we not need to perform this check when updating a cache's config as well?

vblagoje: @ryanemerson right, we do even when we update. Let me look at the createCacheStore method

vblagoje:
@ryanemerson have a look now. Interestingly enough when I throw new Error("text here") it is not caught by deferred but when I just throw "text here" it does. In the latter case, the promise is rejected properly and modal raised for both update/create.

src/services/cache-config/CacheConfigService.ts
```diff
@@ -345,6 +344,25 @@ export class CacheConfigService {
     }
   }
 
+  private transactionHelper(builder: CompositeOpBuilder, address: string[], config: any, create: boolean): void {
+    let validTxMode: boolean = true;
+    if (isNotNullOrUndefined(config.transaction) && isNotNullOrUndefined(config.transaction.TRANSACTION)) {
```
ryanemerson: If `config.transaction` or `config.transaction.TRANSACTION` are null, should we not exist this method and do nothing? Currently if they are null/undefined, we will always go into either the create/update helper

ryanemerson:
@vblagoje Interesting about the exception. From a quick google, it seems like the "proper" way is to passed the deferred instance to the `transactionHelper(..)` function and call reject there. But, I'm very happy with your way if it works as it is easy to follow.

src/services/cache-config/CacheConfigService.ts
vblagoje: Great catch, thanks @ryanemerson

vblagoje:
That was my intuition as well @ryanemerson but we do not have a reference for deferred in update only in create. Let me test all use case scenario one more time!

vblagoje:
@ryanemerson we can fix this issue sort of :-( Say we are creating cache configuration template. We set transaction conf part to trigger an exception. The problem is that when we check for validity of tx node we have already created a cache configuration template on the server. When we fix the tx node we can not execute all the commands again as we have already create the template. We get the duplicate resource error.

vblagoje:
We should somehow make a small framework that navigates the entire conf tree before we send it to server. Here would could add these tx checks, and others in the future, before we send anything to server. Thoughts @ryanemerson? Is it worth doing this? It is not a lot of work but it is the proper solution IMHO

ryanemerson:
@vblagoje Good point! Can we not attempt to delete the template before throwing the error?

vblagoje:
@ryanemerson ok good point. For create op if we get any errors from the server we can delete the template and let the user fix the errors. Memory model stays in the browser. For update op we can not delete the template but only raise the error. Working on this now.

vblagoje:
@ryanemerson things got complicated with options to create cache configuration both as a template and during cache creation. I'll see it could be done by inspecting the tree before we send anything to a server. I know this is not how we do other checks at the moment but IMHO it is how we should do it in the future. I'll leave this issue for Monday.

vblagoje:
@ryanemerson this solution preemptively checks the configuration tree and rejects promise. It is simpler and there is no need to remove configurations. We can add new checks later. The only downside I see is that state of the promise does not seem to be intended to be exposed. However, others are doing what we are doing as well. See https://stackoverflow.com/questions/24091513/get-state-of-angular-deferred for more details.

src/services/cache-config/CacheConfigService.ts
```diff
@@ -303,34 +303,38 @@ export class CacheConfigService {
...
-        this.createHelper(builder, address.concat("transaction", "TRANSACTION"), config.transaction);
-        this.createHelper(builder, address.concat("state-transfer", "STATE_TRANSFER"), config["state-transfer"]);
-        this.createBackupNodes(builder, address.concat("backup"), config.backup);
-
-        if (this.isSecurityAuthorizationEnabled(config)) {
-          this.updateSecurityAuthorization(config);
-          this.createHelper(builder, address.concat("security", "SECURITY"), config.security);
-          this.createHelper(builder, address.concat("security", "SECURITY", "authorization", "AUTHORIZATION"), config.security.SECURITY.authorization);
-        }
+    if (!isRejected(deferred)) {
```
ryanemerson: As we are doing all of the verification before we execute the first post (which makes sense), there is no need to check the status of the promise. Instead, verifyConfiguration can return a string, throw an error, return an object, whatever. This result can then be used to reject the promise outright, or proceed with the DMR steps.

vblagoje: @ryanemerson ok, makes sense. Have a look now. I used try catch because I want any of the verifiers to be able to throw an error. It gets captured and it rejects promise. Otherwise, we have a valid configuration and we proceed with DMR call.

ryanemerson:
Integrated @vblagoje, good work!

PR #218, opened 2017-07-24
Administration console - remote sites are not displayed correctly on …

vblagoje:
Master only. A working proposal to consider as a solution for https://issues.jboss.org/browse/ISPN-7642


vblagoje:
After seeing discussion on JIRA I think this one is GTG @ryanemerson

PR #219, opened 2017-07-26
ISPN-7971 Display a warning that changes in standalone mode are not persisted

vblagoje:
Master only. @ryanemerson @slaskawi this is a suggested solution. We can adjust the detection mechanism for openshift running env in the future. For now, the warning is displayed in standalone mode as well.

### DIFF
--- a/src/module/navbar/NavbarCtrl.ts
+++ b/src/module/navbar/NavbarCtrl.ts
@@ -49,6 +49,10 @@ export class NavbarCtrl {
     return this.authService.isApiAvailable();
   }
 
+  isStandaloneModeOpenshift(): boolean {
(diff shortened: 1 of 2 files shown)


vblagoje:
Here is the suggested warning as a standard Patternfly 
screen shot 2017-07-26 at 9 45 27 am
alert

ryanemerson:
@vblagoje This needs more work, as a standalone server in a non-openshift environment can update a containers configuration. Furthermore, it makes sense for this warning to be displayed when we have clustered standalone instances, as there is currently no way for the updated config to be propagated around the cluster.

vblagoje:
Postponed. See discussion at https://issues.jboss.org/browse/ISPN-7971

PR #220, opened 2017-08-04
Add typedoc documentation

vblagoje:
Adding typedoc documentation. I had to make a change in src/services/domain/IDomain.ts for the typedoc to work. It caught the type assignment mismatch that I fixed. Have a look @ryanemerson

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -40,6 +40,8 @@ gulp.task('serve:dist', ['build:dist'], require('./tasks/server')(gulp, config.d
 gulp.task('serve', ['compile:src', 'less', 'fonts', 'html', 'json', 'assets'],
     require('./tasks/server')(gulp, config.srcDir, config.watchDir, true, config.projectDir, config));
 
+gulp.task('typedoc',require('./tasks/typedoc')(gulp, 'src/services/**/*.ts'));
+
 gulp.task('check:eslint', require('./tasks/check-eslint')(gulp, config));
 gulp.task('check:tslint', ['check:tslint:src']);
 gulp.task('check:tslint:src', require('./tasks/check-tslint')(gulp, config.srcDir, config.tsLintSrcConf));
--- a/src/services/domain/IDomain.ts
+++ b/src/services/domain/IDomain.ts
@@ -1,6 +1,5 @@
-import {IServerAddress} from "../server/IServerAddress";
 
 export interface IDomain {
   controller: string;
-  servers: IServerAddress[]; // Map of format hostname: [server1, server2]
+  servers: string[][]; // Map of format hostname: [server1, server2]
 }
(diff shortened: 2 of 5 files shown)


src/services/domain/IDomain.ts
```diff
@@ -1,6 +1,5 @@
-import {IServerAddress} from "../server/IServerAddress";
 
 export interface IDomain {
   controller: string;
-  servers: IServerAddress[]; // Map of format hostname: [server1, server2]
+  servers: string[][]; // Map of format hostname: [server1, server2]
```
ryanemerson: Why do we need this change?

vblagoje: Typedoc flags it as an incompatible assignment. I tested in a debugger and it indeed is. We are not dealing here with IServerAddress types but with an array of string arrays. Please see it for your self and check the types

gulpfile.js
```diff
@@ -40,6 +41,18 @@ gulp.task('serve:dist', ['build:dist'], require('./tasks/server')(gulp, config.d
...
+    .src(['src/services/**/*.ts'])
+    .pipe(typedoc({
+      module: 'commonjs',
+      target: 'es5',
+      out: 'docs/',
+      name: 'Infinispan management console'
+    }))
+    ;
+});
+
```
ryanemerson: This method should be in its own file like the other gulp tasks. `gulpfile.js` is just an entry point.

gulpfile.js
```diff
@@ -40,6 +41,18 @@ gulp.task('serve:dist', ['build:dist'], require('./tasks/server')(gulp, config.d
 gulp.task('serve', ['compile:src', 'less', 'fonts', 'html', 'json', 'assets'],
     require('./tasks/server')(gulp, config.srcDir, config.watchDir, true, config.projectDir, config));
 
+gulp.task('typedoc', function() {
+  return gulp
+    .src(['src/services/**/*.ts'])
```
ryanemerson: Why don't we just create the docs for all of our src files? That way we have everything

vblagoje: It gets really messy/busy and I am not sure we need those. Service API is the most important part IMHO. I am open to including if there is a decent argument for it.

gulpfile.js
vblagoje: Ok +1

gulpfile.js
ryanemerson: No argument, I was just curious

vblagoje:
Have a look now @ryanemerson

ryanemerson:
Integrated https://github.com/infinispan/infinispan-management-console/commit/ec695b2ff72b824aa193348d5d644bb2574fceab, thanks @vblagoje

Btw, I noticed that we didn't clean the docs dir so I added [this](https://github.com/infinispan/infinispan-management-console/blob/ec695b2ff72b824aa193348d5d644bb2574fceab/tasks/clean.js#L29)

PR #221, opened 2017-09-11
ISPN-8279 Console support for Off-Heap

vblagoje:
Master only. Preview of off-heap memory stats for both cache and container(server node). @ryanemerson in order to test this PR please create a cache configuration template with off-heap memory. Make sure to set type to "OFF-HEAP", eviction to "MEMORY" or "COUNT" and then create a cache C using this off heap template. Perhaps create one template with "MEMORY" and the other with "COUNT" eviction. Then go on cache C stats screen, you should see this:
screen shot 2017-09-12 at 4 32 11 am

Not the new card displaying stats for off heap memory usage. We are able to display max memory used only if we have "OFF-HEAP" with eviction set to "MEMORY". Otherwise, we only display off-heap memory used. Please check that "Off heap" card is not shown for other caches as well. Also please check that "Nodes" tab displays "Off heap memory used".

On single server page stats you should also find Off heap section. 
screen shot 2017-09-12 at 4 35 03 am

Here we do not have max off heap reading as it is a bit more complicated to calculate and it is not available in DMR. I left that part for later not to complicate things too much.

### DIFF
--- a/src/services/cache/CacheService.ts
+++ b/src/services/cache/CacheService.ts
@@ -72,7 +72,9 @@ export class CacheService {
     let request: IDmrRequest = {
       address: this.generateContainerAddress(container, profile).concat(type, name)
     };
-    this.dmrService.readResource(request).then((response) => deferred.resolve(new Cache(name, type, response.configuration)));
+    this.dmrService.readResource(request).then((response) => {
+      deferred.resolve(new Cache(name, type, response.configuration));
+    }, error => deferred.reject(error));
     return deferred.promise;
   }
 
@@ -87,7 +89,7 @@ export class CacheService {
         c.configModel = response;
         deferred.resolve(c);
       });
-    });
+    }, error => deferred.reject(error));
     return deferred.promise;
   }
 
--- a/src/services/cache/CacheService.ts
+++ b/src/services/cache/CacheService.ts
@@ -76,6 +76,21 @@ export class CacheService {
     return deferred.promise;
   }
 
+  getCacheWithConfiguration(container: ICacheContainer, type: string, name: string): ng.IPromise<ICache> {
+    let deferred: ng.IDeferred<ICache> = this.$q.defer<ICache>();
+    let request: IDmrRequest = {
+      address: this.generateContainerAddress(container.name, container.profile).concat(type, name),
+    };
+    this.dmrService.readResource(request).then((response) => {
+      let c: ICache = new Cache(name, type, response.configuration);
+      this.getCacheTemplate(container, type, name).then((response) => {
+        c.configModel = response;
+        deferred.resolve(c);
+      });
+    });
+    return deferred.promise;
+  }
+
   getCacheTemplate(container: ICacheContainer, type: string, name: string): ng.IPromise<any> {
     let deferred: ng.IDeferred<any> = this.$q.defer();
     this.getCache(name, type, container.name, container.profile)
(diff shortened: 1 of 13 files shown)


src/services/cache/CacheService.ts
```diff
@@ -76,6 +76,21 @@ export class CacheService {
...
+    let request: IDmrRequest = {
+      address: this.generateContainerAddress(container.name, container.profile).concat(type, name),
+    };
+    this.dmrService.readResource(request).then((response) => {
+      let c: ICache = new Cache(name, type, response.configuration);
+      this.getCacheTemplate(container, type, name).then((response) => {
+        c.configModel = response;
+        deferred.resolve(c);
+      });
+    });
```
ryanemerson: We should re-use the `getCache` method

```typescript
getCacheWithConfiguration(container: ICacheContainer, type: string, name: string): ng.IPromise<ICache> {
    let deferred: ng.IDeferred<ICache> = this.$q.defer<ICache>();
    this.getCache(name, type, container.name, container.profile)
      .then(cache => {
        this.getCacheTemplate(container, type, name)
          .then(template => {
            cache.configModel = template;
            deferred.resolve(cache);
          })
      });
    return deferred.promise;
  }
```

We should have the scenarios where any of these operations fail as well

vblagoje: +1 I know we should re-use getCache but it getCacheTemplate calls getCache and then I would introduce recursion. For such a small fix I would not want to risk potential problems for a PR that needs to go downstream in 7.2. We can revisit all these cache service methods post release. I'll add deferred rejecting for errors.

vblagoje:
Done, have a look EOD today if you can @ryanemerson Let's integrate it by tomorrow!

src/services/launchtype/LaunchTypeService.ts
```diff
@@ -78,6 +78,10 @@ export class LaunchTypeService {
     }
   }
 
+  getMemoryUnit(): MemoryUnits {
+    return MemoryUnits.MB;
```
ryanemerson: I really like this solution, but I don't understand why the memory units will be influenced by the LanchType? As `convertBytes` always assumes that the input is bytes, we can inspect the passed value and determine the appropriate units to convert to.

src/services/cache/CacheService.ts
ryanemerson: Whoops, I didn't realise that getCacheTemplate called `getCache` good spot :+1:

src/services/launchtype/LaunchTypeService.ts
vblagoje: Hahaha, I did not know where else to put convertBytes. But how do you know appropriate units to convert to? Ideally, it would be user setting!

ryanemerson: So if `bytes >= (1024^1) && bytes < (1024^2)` = KB` etc

vblagoje: Should I just put it in Util for now?

ryanemerson: Well if it's always going to be MB then there is no need to specify the Memory unit as part of the `convertBytes` function, instead just return `round((bytes / 1024 / 1024), 2) + " MB";` everytime. That being said, the code to figure out whether to treat as KB, MB, GB is trivial

vblagoje: Ok, I would go with automatic version then. If bytes >= (1024^1) && bytes < (1024^2) use KB and so on. I'll do that!

vblagoje:
@ryanemerson I made the automatic unit conversion as we talked. I've tested it with hard coded values to make sure it converts properly to Bytes, KB, MB, and GB.

vblagoje:
Actually @ryanemerson I need to make one more change! One sec

vblagoje:
Ok, should be ready now @ryanemerson . Please squash all commits and then let's see if we can see everything being GTG in totality.

ryanemerson:
@vblagoje Integrated https://github.com/infinispan/infinispan-management-console/commit/669f20d486ff3ce2b5663e85832d10ce679d2bf5, thanks!

PR #225, opened 2017-11-20
ISPN-7927 User Interface to manage Cluster Counters

vblagoje:
Master only as it requires 9.2 counters DMR. PR for counters is at https://github.com/infinispan/infinispan/pull/5570 and as soon as it gets integrated we can try out this PR and integrate it as well. In the meantime, let's look at things that need to be ironed out so this PR is ready to go as well.

### DIFF
--- a/src/module/cache-container/CacheContainer.ts
+++ b/src/module/cache-container/CacheContainer.ts
@@ -10,6 +10,7 @@ import {IRedirectState} from "../../common/IRedirectState";
 import {ICacheContainer} from "../../services/container/ICacheContainer";
 import {ContainerService} from "../../services/container/ContainerService";
 import {IStateParamsService} from "angular-ui-router";
+import {CountersCtrl} from "./counters/CountersCtrl";
 
 const module: ng.IModule = App.module("managementConsole.cache-container", []);
 
@@ -50,6 +51,18 @@ module.config(($stateProvider: ng.ui.IStateProvider) => {
       }],
       templates: ["container", "cacheConfigService", (container, cacheConfigService) => {
         return cacheConfigService.getAllContainerTemplatesShallow(container);
+      }]
+    }
+  });
+
+  $stateProvider.state("container.counters", {
+    url: "/counters",
+    templateUrl: "module/cache-container/counters/view/counters.html",
+    controller: CountersCtrl,
+    controllerAs: "counterCtrl",
+    resolve: {
+      counters: ["container", "counterService", (container, counterService) => {
+        return counterService.getAllCounters(container);
       }]
     }
   });
--- a/src/module/cache-container/counters/CountersCtrl.ts
+++ b/src/module/cache-container/counters/CountersCtrl.ts
@@ -1,0 +1,59 @@
+import {ICacheContainer} from "../../../services/container/ICacheContainer";
+import IModalService = angular.ui.bootstrap.IModalService;
+import {ICounter} from "../../../services/counters/ICounter";
+import {CounterCreateModalCtrl} from "./CounterCreateModalCtrl";
+import {CounterService} from "../../../services/counters/CounterService";
+import {openConfirmationModal} from "../../../common/dialogs/Modals";
+import {IStateService} from "angular-ui-router";
+import {StrongCounter} from "../../../services/counters/StrongCounter";
+import {isNotNullOrUndefined} from "../../../common/utils/Utils";
+
+export class CountersCtrl {
+  static $inject: string[] = ["$state", "$uibModal", "container", "counterService", "counters"];
+
+  constructor(private $state: IStateService,
+              private $uibModal: IModalService,
+              private container: ICacheContainer,
+              private counterService: CounterService,
+              private counters: ICounter []) {
+  }
+
+  createCounterModal(): void {
+    this.$uibModal.open({
+      templateUrl: "module/cache-container/counters/view/counters-create.html",
+      controller: CounterCreateModalCtrl,
+      controllerAs: "modalCtrl",
+      resolve: {
+        container: (): ICacheContainer => this.container,
+        counterService: (): CounterService => this.counterService
+      }
+    });
+  }
+
+  reset(counter: ICounter): void {
+    let message: string = "Reset counter '" + counter.getName() + "'?";
+    openConfirmationModal(this.$uibModal, message, "pficon")
+      .result
+      .then(() => this.counterService.reset(this.container, counter))
+      .then(() => this.$state.reload());
+  }
+
+  remove(counter: ICounter): void {
+
+    let message: string = "Remove counter '" + counter.getName() + "'?";
+    openConfirmationModal(this.$uibModal, message, "pficon pficon-delete")
+      .result
+      .then(() => this.counterService.remove(this.container, counter))
+      .then(() => this.$state.reload());
+  }
+
+  getBounds(counter: ICounter): string {
+    if (counter instanceof StrongCounter) {
+      let strongCounter: StrongCounter = <StrongCounter> counter;
+      let lb: number = strongCounter.getLowerBound();
+      let ub: number = strongCounter.getUpperBound();
+      return "[" + (isNotNullOrUndefined(lb) ? lb : "") + "," + (isNotNullOrUndefined(ub) ? ub : "") + "]";
+    }
+    return "N/A";
+  }
+}
--- a/src/module/cache-container/counters/view/counters.html
+++ b/src/module/cache-container/counters/view/counters.html
@@ -1,0 +1,44 @@
+<div class="row">
+  <div class="col-sm-12 col-md-12">
+    <table class="datatable table table-striped table-bordered">
+      <thead>
+      <tr>
+        <th colspan="7" class="text-right">
+          <div class="btn-group" role="group">
+            <a role="button" class="btn btn-default" data-toggle="modal" ng-click="counterCtrl.createCounterModal()"><span
+              class="pficon pficon-add-circle-o"></span> Create new counter</a>
+          </div>
+
+        </th>
+      </tr>
+      <tr>
+        <th>Name</th>
+        <th>Type</th>
+        <th>Storage</th>
+        <th>Initial value</th>
+        <th>Current value</th>
+        <th>Bounds</th>
+        <th width="170" class="text-center">Actions</th>
+      </tr>
+      </thead>
+      <tbody>
+      <tr ng-repeat="counter in counterCtrl.counters">
+        <td>{{counter.getName()}}</td>
+        <td>{{counter.toString()}}</td>
+        <td>{{counter.getStorage()}}</td>
+        <td>{{counter.getInitialValue()}}</td>
+        <td>{{counter.getCurrentValue()}}</td>
+        <td>{{counterCtrl.getBounds(counter)}}</td>
+        <td class="text-center">
+          <div class="btn-group">
+            <button type="button" class="btn btn-default" ng-click="counterCtrl.reset(counter)"><i class="pficon pficon-edit"></i> Reset</button>
+          </div>
+          <div class="btn-group">
+            <button type="button" class="btn btn-default" ng-click="counterCtrl.remove(counter)"><i class="pficon pficon-delete"></i> Remove</button>
+          </div>
+        </td>
+      </tr>
+      </tbody>
+    </table>
+  </div>
+</div>
--- a/src/services/counters/CounterService.ts
+++ b/src/services/counters/CounterService.ts
@@ -1,0 +1,166 @@
+import {App} from "../../ManagementConsole";
+import {DmrService} from "../dmr/DmrService";
+import {ICounter} from "./ICounter";
+import {IDmrRequest} from "../dmr/IDmrRequest";
+import {SocketBindingService} from "../socket-binding/SocketBindingService";
+import IQService = angular.IQService;
+import {LaunchTypeService} from "../launchtype/LaunchTypeService";
+import {isNotNullOrUndefined, traverse, deepValue} from "../../common/utils/Utils";
+import {ICacheContainer} from "../container/ICacheContainer";
+import {JGroupsService} from "../jgroups/JGroupsService";
+import {IServerAddress} from "../server/IServerAddress";
+import {WeakCounter} from "./WeakCounter";
+import {StrongCounter} from "./StrongCounter";
+
+const module: ng.IModule = App.module("managementConsole.services.counter", []);
+
+export const STRONG_COUNTER: string = "STRONG";
+export const WEAK_COUNTER: string = "WEAK";
+export const COUNTER_TYPES: string [] = ["STRONG", "WEAK"];
+export const COUNTER_STORAGE_TYPES: string [] = ["PERSISTENT", "VOLATILE"];
+
+export class CounterService {
+  static $inject: string[] = ["$q", "jGroupsService", "dmrService", "socketBindingService", "launchType"];
+
+  constructor(private $q: IQService, private jGroupsService: JGroupsService, private dmrService: DmrService,
+              private socketBindingService: SocketBindingService, private launchType: LaunchTypeService) {
+  }
+
+  doCountersExist(cacheContainer: string, profile: string): ng.IPromise<boolean> {
+    let deferred: ng.IDeferred<boolean> = this.$q.defer<boolean>();
+    let request: IDmrRequest = {
+      address: this.getCountersConfigurationAddress(cacheContainer, profile)
+    };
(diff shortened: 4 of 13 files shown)


vblagoje:
@ryanemerson please do not look at this one yet, I'll polish it a bit more tonight.

vblagoje:
@ryanemerson could you give this PR a look today?

src/module/cache-container/counters/CountersCtrl.ts
```diff
@@ -0,0 +1,44 @@
...
+      controllerAs: "modalCtrl",
+      resolve: {
+        container: (): ICacheContainer => this.container,
+        counterService: (): CounterService => this.counterService
+      }
+    });
+  }
+
+  reset(counter: ICounter): void {
+    this.counterService.reset(this.container, counter).then(() => this.$state.reload());
```
ryanemerson: This should probably be in a confirmation modal

src/services/counters/Counter.ts
```diff
@@ -0,0 +1,63 @@
...
+    return this.type;
+  }
+
+  isStrong(): boolean {
+    return this.getType() === Counter.STRONG_TYPE;
+  }
+
+  isWeak(): boolean {
+    return !this.isStrong();
+  }
```
ryanemerson: This seems like a case for inheritance. `AbstractCounter`, `StrongCounter extends AbstractCounter` etc.

src/module/cache-container/counters/CountersCtrl.ts
vblagoje: I thought about it as well but then I remembered Heather telling me how only resource create/destruct and operations that leave the system in a different state should be confirmed to avoid modal overclicking. It is fine art to decide this one...

src/services/counters/Counter.ts
vblagoje: Right, good call, let me see how it looks and I'll update the PR

src/module/cache-container/counters/CountersCtrl.ts
ryanemerson: Surely resetting a counter will leave the system in a different state?

vblagoje: But then what do we do about cache statistic resets that do not have modal. Well, not the same thing...ok I see. You are right!

vblagoje:
@ryanemerson updated as we agreed.

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  createStrongCounter(name: string, storage: string, initialValue: number, currentValue: number, lowerBound:number, upperBound: number, dmr: any): ICounter {
+    return new StrongCounter(name, storage, initialValue, currentValue, lowerBound, upperBound, dmr);
+  }
+
+  createWeakCounter(name: string, storage: string, initialValue: number, currentValue: number, concurrency: number, dmr: any): ICounter {
+    return new WeakCounter(name, storage, initialValue, currentValue, concurrency, dmr);
+  }
+
+  getAllCounters(cacheContainer: ICacheContainer): ng.IPromise<ICounter[]> {
```
ryanemerson: We can create a method `doCountersExist` that returns true if a counter resource exists. This would be a simple read-resource operation on the correct DMR path, if it isn't successful, then no counters exist. You can then wrap `getAllCounters` with ^ methid. So if counters exist, `getAllCounters`, else return an empty array. 

You can then use `doCountersExist` to enable/disable the counter pages and/or tabs

vblagoje: @ryanemerson have a look now please

src/services/counters/StrongCounter.ts
```diff
@@ -0,0 +1,34 @@
+import {isNotNullOrUndefined} from "../../common/utils/Utils";
+import {AbstractCounter} from "./AbstractCounter";
+export class StrongCounter extends AbstractCounter {
+
+  constructor(protected name: string, protected storage: string,
+              protected initialValue: number, protected currentValue: number,
+              protected lowerBound: number, protected upperBound: number, protected dmr: any) {
```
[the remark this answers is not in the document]
vblagoje: What exactly do you mean here?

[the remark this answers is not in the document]
vblagoje: Aha I see

src/services/counters/WeakCounter.ts
```diff
@@ -0,0 +1,33 @@
...
+  getConcurrency(): number {
+    return this.concurrency;
+  }
+
+  getLowerBound(): number {
+    return undefined;
+  }
+
+  getUpperBound(): number {
+    return undefined;
```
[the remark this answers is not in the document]
vblagoje: Ok, this is more of a philosophical difference in opinion, no? I like dealing with ICounter interface rather than check types, it simplifies view code...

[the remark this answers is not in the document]
ryanemerson: Well if you are going to go down inheritance route to reduce code repetition then the code how it currently is doesn't really make sense. However, if you would prefer just to have implementations of `ICounter.ts` and have some unnecessary code duplication I can live with that, you make the call. But, it doesn't make sense to have a strange hybrid where we use some inheritance but duplicate most of the code anyway. 

Furthermore, you actually only call `hasBounds` in one view, so you're writing a lot of boilerplate code to avoid `instanceof StrongCounter`. With regards to `isStrong()` the only places you use it in the view can be simplified by adding a `toString` method to `IContainer` and providing STRONG/WEAK as the return value in the appropriate impl.

src/module/cache-container/counters/view/counters.html
```diff
@@ -0,0 +1,46 @@
...
+        <th>Current value</th>
+        <th>Bounds</th>
+        <th width="170" class="text-center">Actions</th>
+      </tr>
+      </thead>
+      <tbody>
+      <tr ng-repeat="counter in counterCtrl.counters">
+        <td>{{counter.getName()}}</td>
+        <td ng-if="counter.isStrong()">STRONG</td>
+        <td ng-if="counter.isWeak()">WEAK</td>
```
ryanemerson: By adding `toString` to the `ICounter` interface we can replace these two lines with `<td>{{counter.toString()}}</td>`

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,176 @@
...
+    let request: IDmrRequest = {
+      address: counterAddress
+    };
+    this.dmrService.readResource(request).then(r => {
+      deferred.resolve(true);
+    }).catch(e => {
+      deferred.resolve(false);
+    });
+    return deferred.promise;
+  }
```
ryanemerson: Can be simplified to:

```typescript
  doCountersExist(cacheContainer: string, profile: string): ng.IPromise<boolean> {
    let deferred: ng.IDeferred<boolean> = this.$q.defer<boolean>();
    let request: IDmrRequest = {
      address: this.getCountersConfigurationAddress(cacheContainer, profile)
    };
    this.dmrService.readResource(request).then(r => deferred.resolve(true), e => deferred.resolve(false));
    return deferred.promise;
  }
```

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,176 @@
...
+    let deferred: ng.IDeferred<ICounter[]> = this.$q.defer<ICounter[]>();
+    this.jgroupsService.getServerGroupCoordinator(cacheContainer.serverGroup).then((coord: IServerAddress) => {
+
+      let request: IDmrRequest = <IDmrRequest>{
+        address: this.getCountersRuntimeAddress(cacheContainer.name, coord.host, coord.name),
+        "recursive-depth": 2,
+        "include-runtime": true
+      };
+
+      this.dmrService.readResource(request)
```
ryanemerson: Will this not fail if `counters=COUNTERS` hasn't been created?

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,176 @@
...
+    return deferred.promise;
+  }
+
+  update(counter: ICounter, profile: string, container: string): ng.IPromise<any> {
+    return this.save(counter, profile, container, ["is-new-node"]);
+  }
+
+  create(counter: ICounter, profile: string, container: string): ng.IPromise<any> {
+    counter["is-new-node"] = true;
+    return this.doCountersExist(container, profile).then((countersExists: boolean) => {
```
ryanemerson: this can be simplified to `then(countersExist => ...`

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,176 @@
...
+      if (countersExists) {
+        return this.createHelper(counter, profile, container);
+      } else {
+        return this.createCountersDMRPath(container, profile).then((created: boolean) => {
```
ryanemerson: this can be simplified to `then(created => ...`

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,176 @@
...
+          if (created) {
+            return this.createHelper(counter, profile, container);
+          }
+        })
```
ryanemerson: missing `;`

src/services/counters/CounterService.ts
ryanemerson: We need to properly handle the failure of any of these promises, that way an error message will be displayed on the modal

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,176 @@
...
+    let deferred: ng.IDeferred<boolean> = this.$q.defer<boolean>();
+    let counterAddress: string [] = this.getCountersConfigurationAddress(cacheContainer, profile);
+    let request: IDmrRequest = {
+      address: counterAddress
+    };
+    this.dmrService.add(request).then(r => {
+      deferred.resolve(true);
+    }).catch(e => {
+      deferred.resolve(false);
+    });
```
ryanemerson: There is no reason for this method to return `IPromise<boolean>` as we never use the return value. Instead we can return IPromise<void> and then have the whole method as:

```typescript
return this.dmrService.add({
      address: this.getCountersConfigurationAddress(cacheContainer, profile)
    });
```

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,176 @@
...
+          if (created) {
+            return this.createHelper(counter, profile, container);
+          }
+        })
+      }
+    });
+  }
+
+  createHelper(counter: ICounter, profile: string, container: string): ng.IPromise<any> {
+    if (this.doCountersExist(container, profile)) {
```
ryanemerson: No need to check that the counters exists again as this another server call and `createHelper` is only called once we know that counters do exist. We should make this method private to ensure that it is used correctly

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,176 @@
...
+    return this.dmrService.executePost({
+      address: address,
+      operation: "remove"
+    });
+  }
+
+  save(counter: ICounter, profile: string, container: string, excludedAttributes: string []): ng.IPromise<any> {
+    let builder: CompositeOpBuilder = new CompositeOpBuilder();
+    let dmrAddress: string [] = this.getCounterAddress(counter, container, profile);
+    this.dmrService.traverseDMRTree(builder, counter, dmrAddress, excludedAttributes);
```
ryanemerson: I'm lost at this point. It's only a single add operation to create a counter. Why do we need to do all of this traversal? Can't we just keep it simple and do a simple add operation? e.g. https://github.com/infinispan/infinispan-management-console/blob/master/src/services/server/ServerService.ts#L30-L38

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,176 @@
...
+      if (counter.isStrong()) {
+        return this.save(counter, profile, container, ["is-new-node", "type", "concurrency"]);
+      } else {
+        return this.save(counter, profile, container, ["is-new-node", "type", "lower-bound", "upper-bound"]);
+      }
+    }
+  }
+
+  reset(container: ICacheContainer, counter: ICounter): ng.IPromise<any> {
+    let address: string[] = this.getCounterAddress(counter, container.name, container.profile);
```
ryanemerson: No need for this variable

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,176 @@
...
+    return this.dmrService.executePost({
+      address: address,
+      operation: "counter-reset"
+    });
+  }
+
+  remove(container: ICacheContainer, counter: ICounter): ng.IPromise<any> {
+    let address: string[] = this.getCounterAddress(counter, container.name, container.profile);
```
ryanemerson: No need for this variable

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  private getCounterAddress(c: ICounter, container: string, profile?: string): string[] {
+    let address: string [] = this.getCountersConfigurationAddress(container, profile);
+    address = address.concat(c.isStrong() ? "strong-counter" : "weak-counter").concat(c.getName());
+    return address;
+  }
+
+  private getContainerAddress(container: string, preappendPath: string []): string[] {
+    let address: string[] = preappendPath;
```
ryanemerson: No need for this variable

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,176 @@
...
+          }, trail);
+          return counters;
+        }).then(counters => {
+        deferred.resolve(counters);
+      })
+    });
+    return deferred.promise;
+  }
+
+  update(counter: ICounter, profile: string, container: string): ng.IPromise<any> {
```
ryanemerson: AFAICT this is never called

src/services/counters/CounterService.ts
vblagoje: Ok, I agree but I wanted to use this op that already creates all the necessary DMR operations for sava. Can we leave this for rework later?

ryanemerson: Ideally not as this is potentially creating more ops then required and it is something that will get forgot about. It should be a simple implementation, something like:

```typescript
createCounter(counter: ICounter, profile: string, container: string): ng.IPromise<void> {
    let request: IDmrRequest;
    if (counter instanceof StrongCounter) {
      let strongCounter: StrongCounter = <StrongCounter> counter;
      request = <IDmrRequest> {
        address: this.getCounterAddress(counter, container, profile),
        "initial-value": strongCounter.getInitialValue(),
        ...
      }
    } else {
      request = <IDmrRequest> {
        ...
      };
    }
    return this.dmrService.add(request);
  }
```

src/module/cache-container/CacheContainer.ts
```diff
@@ -54,6 +55,24 @@ module.config(($stateProvider: ng.ui.IStateProvider) => {
...
+  $stateProvider.state("container.counters", {
+    url: "/counters",
+    templateUrl: "module/cache-container/counters/view/counters.html",
+    controller: CountersCtrl,
+    controllerAs: "counterCtrl",
+    resolve: {
+      counters: ["container", "counterService", (container, counterService) => {
+        return counterService.doCountersExist(container.name, container.profile).then(countersExist => {
+          if (countersExist) {
+            return counterService.getAllCounters(container);
```
ryanemerson: Ideally `doCountersExist` would be part of `getAllCounters, with [] returned when counters don't exist, and that way the logic remains in the service.

src/services/counters/CounterService.ts
ryanemerson: ^ I guess not because it's guarded in `CacheContainer.ts` but we should really have this check in the service

src/services/counters/CounterService.ts
```diff
@@ -0,0 +1,156 @@
...
+
+const module: ng.IModule = App.module("managementConsole.services.counter", []);
+
+export class CounterService {
+  static $inject: string[] = ["$q", "jGroupsService", "dmrService", "socketBindingService", "launchType"];
+
+  public static STRONG_COUNTER: string = "STRONG";
+  public static WEAK_COUNTER: string = "WEAK";
+  public static COUNTER_TYPES: string [] = ["STRONG", "WEAK"];
+  public static COUNTER_STORAGE_TYPES: string [] = ["PERSISTENT", "VOLATILE"];
```
ryanemerson: You've been spending too much time in Java land :smile:. We should export these constants as `const` like we have in other services https://github.com/infinispan/infinispan-management-console/blob/master/src/services/cache-config/CacheConfigService.ts#L21

src/module/cache-container/CacheContainer.ts
vblagoje: I did this

src/services/counters/CounterService.ts
vblagoje: True, good catch

PR #226, opened 2017-12-22
ISPN-8642 Administration console - updating scheme doesn't close the update dialog

vblagoje:
@ryanemerson I simplied the workflow and error feedback to adhere the already established Patternfly paradigms. We should in fact go once again through all modal dialogs and update any inconsistencies like this one.

### DIFF
--- a/src/module/cache-container/config/schemas/SchemaModalCtrl.ts
+++ b/src/module/cache-container/config/schemas/SchemaModalCtrl.ts
@@ -30,9 +30,7 @@ export class SchemaModalCtrl {
     this.schemaService.registerProtoSchema(this.container, this.schema)
       .then(() => {
           this.successfulOperation = true;
-          if (!this.editing) {
-            this.$uibModalInstance.close(true);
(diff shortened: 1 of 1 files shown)


src/module/cache-container/config/schemas/SchemaModalCtrl.ts
```diff
@@ -3,16 +3,16 @@ import {ISchemaDefinition} from "../../../../services/schemas/ISchemaDefinition"
 import {SchemaService} from "../../../../services/schemas/SchemasService";
 import {stringEndsWith} from "../../../../common/utils/Utils";
 import IModalServiceInstance = angular.ui.bootstrap.IModalServiceInstance;
```
ryanemerson: This is now unused, see `gulp check`

vblagoje: Yes, we are definitely doing something wrong here. We should not remove the resource on the failed update.

PR #233, opened 2018-03-03
ISPN-8827 Disable browser cache for some resources in Admin Console

vblagoje:
Master and preview only. Please have a look, particularly about dist/index.html file as during every build we'll have that file dirty due to string replacement. We could move index.html to index_tmp.html, keep it versioned, and then during build make index.html and gitignore it.

### DIFF
--- a/tasks/clean.js
+++ b/tasks/clean.js
@@ -13,8 +13,7 @@ module.exports = (gulp, dir, config, cleanAll) => () => {
 
   var cleanDist = function (directory) {
     return del([
-      path.join(directory, '**', '*'),
-      '!' + path.join(directory, 'index.html')
+      path.join(directory, '**', '*')
     ]);
   };
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 node
 node_modules
 dist
-!dist/index.html
 .idea
 .vscode
 .tmp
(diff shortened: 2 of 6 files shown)


ryanemerson:
@vblagoje Looks good! I like the idea of only storing `index_tmp.html` in git and then transforming that into `index.html`. This would also allow us to remove the `dist` dir from git entirely and we can store the file in `src` where it belongs.

tasks/clean.js
```diff
@@ -13,8 +13,7 @@ module.exports = (gulp, dir, config, cleanAll) => () => {
 
   var cleanDist = function (directory) {
     return del([
-      path.join(directory, '**', '*'),
-      '!' + path.join(directory, 'index.html')
+      path.join(directory, '**', '*')
```
vblagoje: @ryanemerson I think this step is necessary to remove index.html on clean. Please confirm

vblagoje:
Thanks @ryanemerson It should be good to go now. But please confirm by running yourself. Thanks

vblagoje:
@ryanemerson while I did not explicitly remove /dist from git I believe it is automatically removed after the only file in that directory (i.e. index.html) is removed. I tried it with switching branches where dist exists and this branch. Try it as well yourself.

PR #235, opened 2018-03-26
ISPN-8759 - Administration console - add data manipulation capabilities

vblagoje:
Preview only. More UI work coming after we receive feedback from the team. See https://youtu.be/AE82JH_QDQw for more details

### DIFF
--- a/src/components/query-panel/QueryPanelCtrl.ts
+++ b/src/components/query-panel/QueryPanelCtrl.ts
@@ -1,6 +1,7 @@
 import CodeMirror = require("codemirror");
 import {RestService} from "../../services/rest/RestService";
 import {EditorConfiguration} from "codemirror";
+import {isNotNullOrUndefined, isObject, isEmptyString} from "../../common/utils/Utils";
 
 export class QueryPanelCtrl implements ng.IComponentController {
   static $inject: string[] = ["restService", "$timeout"];
@@ -11,19 +12,28 @@ export class QueryPanelCtrl implements ng.IComponentController {
   statusCode: number;
   editor: CodeMirror.Editor;
   panelType:string;
+  inputParams: string [];
 
   constructor(private restService: RestService, private $timeout: ng.ITimeoutService) {
   }
 
   executeAction(): void {
-    let inputParam1: string = angular.element(document.querySelector("#" + this.name + "-input-text")).val();
-    let inputParam2: string = angular.element(document.querySelector("#" + this.name + "-input-text-area")).val();
-    this.executionActionHelper(this.name, inputParam1, inputParam2).then((response: any) => {
+    this.executionActionHelper(this.name, this.inputParams[0], this.inputParams[1]).then((response: any) => {
       this.statusCode = 200;
-      this.editor.setValue(JSON.stringify(response.data, null, "\t"));
+      if (isNotNullOrUndefined(response) && isNotNullOrUndefined(response.data)) {
+        if (isObject(response.data)) {
+          this.editor.setValue(JSON.stringify(response.data, null, "\t"));
+        } else {
+          this.editor.setValue("Request executed successfully. No response returned.");
+        }
+      }
     }).catch((error: any) => {
       this.statusCode = 401;
-      this.editor.setValue(JSON.stringify(error, null, "\t"));
+      if (isEmptyString(error)) {
+        this.editor.setValue("Error response with no further details returned");
+      } else {
+        this.editor.setValue(JSON.stringify(error, null, "\t"));
+      }
     });
   }
 
--- a/src/components/query-panel/QueryPanelCtrl.ts
+++ b/src/components/query-panel/QueryPanelCtrl.ts
@@ -18,7 +18,7 @@ export class QueryPanelCtrl implements ng.IComponentController {
   }
 
   executeAction(): void {
-    this.executionActionHelper(this.name, this.inputParams[0], this.inputParams[1]).then((response: any) => {
+    this.executionActionHelper().then((response: any) => {
       this.statusCode = 200;
       if (isNotNullOrUndefined(response) && isNotNullOrUndefined(response.data)) {
         if (isObject(response.data)) {
@@ -37,26 +37,26 @@ export class QueryPanelCtrl implements ng.IComponentController {
     });
   }
 
-  executionActionHelper(type: string, param1: string, param2: any): ng.IPromise<any> {
+  executionActionHelper(): ng.IPromise<any> {
     let response: ng.IPromise<any>;
-    switch (type) {
+    switch (this.name) {
       case "SEARCH":
-        response = this.restService.executeCacheQuery(this.cacheName, param2);
+        response = this.restService.executeCacheQuery(this.cacheName, this.inputParams[1]);
         break;
       case "POST":
-        response = this.restService.executeCachePost(this.cacheName, param1, param2);
+        response = this.restService.executeCachePost(this.cacheName, this.inputParams[0], this.inputParams[1]);
         break;
       case "PUT":
-        response = this.restService.executeCachePut(this.cacheName, param1, param2);
+        response = this.restService.executeCachePut(this.cacheName, this.inputParams[0], this.inputParams[1]);
         break;
       case "GET":
-        response = this.restService.executeCacheGet(this.cacheName, param1);
+        response = this.restService.executeCacheGet(this.cacheName, this.inputParams[0]);
         break;
       case "DELETE":
-        response = this.restService.executeCacheDeleteKey(this.cacheName, param1);
+        response = this.restService.executeCacheDeleteKey(this.cacheName, this.inputParams[0]);
         break;
       default:
-        console.log("Unknown request type " + type);
+        console.log("Unknown request type " + this.name);
     }
     return response;
   }
--- a/src/components/query-panel/QueryPanelCtrl.ts
+++ b/src/components/query-panel/QueryPanelCtrl.ts
@@ -1,0 +1,74 @@
+import CodeMirror = require("codemirror");
+import {RestService} from "../../services/rest/RestService";
+import {EditorConfiguration} from "codemirror";
+
+export class QueryPanelCtrl implements ng.IComponentController {
+  static $inject: string[] = ["restService", "$timeout"];
+
+  cacheName: string;
+  name: string;
+  description: string;
+  statusCode: number;
+  editor: CodeMirror.Editor;
+  panelType:string;
+
+  constructor(private restService: RestService, private $timeout: ng.ITimeoutService) {
+  }
+
+  executeAction(): void {
+    let inputParam1: string = angular.element(document.querySelector("#" + this.name + "-input-text")).val();
+    let inputParam2: string = angular.element(document.querySelector("#" + this.name + "-input-text-area")).val();
+    this.executionActionHelper(this.name, inputParam1, inputParam2).then((response: any) => {
+      this.statusCode = 200;
+      this.editor.setValue(JSON.stringify(response.data, null, "\t"));
+    }).catch((error: any) => {
+      this.statusCode = 401;
+      this.editor.setValue(JSON.stringify(error, null, "\t"));
+    });
+  }
+
+  executionActionHelper(type: string, param1: string, param2: any): ng.IPromise<any> {
+    let response: ng.IPromise<any>;
+    switch (type) {
+      case "SEARCH":
+        response = this.restService.executeCacheQuery(this.cacheName, param2);
+        break;
+      case "POST":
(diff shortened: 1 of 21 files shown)


src/components/query-panel/view/delete-panel.html
```diff
@@ -0,0 +1,27 @@
...
+    <uib-accordion-heading>{{ctrl.name}} {{ctrl.description}}</uib-accordion-heading>
+    <div class="row">
+      <div class="col-sm-3 col-lg-2">
+        <div class="form-group">
+          <span>Enter key: </span>
+          <input class="form-control" type="text" id="{{ctrl.name}}-input-text"/>
+        </div>
+      </div>
+      <div class="col-sm-9 col-lg-6">
+        <button type="button" class="btn btn-default execute-button" ng-click="ctrl.executeAction()">Execute</button>
```
ryanemerson: I may be missing something but it seems like the `*-panel.html` files are essentially the same. Can we not just have one generic html template for all operations?

src/components/query-panel/QueryPanelCtrl.ts
```diff
@@ -0,0 +1,74 @@
...
+  statusCode: number;
+  editor: CodeMirror.Editor;
+  panelType:string;
+
+  constructor(private restService: RestService, private $timeout: ng.ITimeoutService) {
+  }
+
+  executeAction(): void {
+    let inputParam1: string = angular.element(document.querySelector("#" + this.name + "-input-text")).val();
+    let inputParam2: string = angular.element(document.querySelector("#" + this.name + "-input-text-area")).val();
```
ryanemerson: Is this not an anti-pattern? We should be passing the input values to the QueryPanel component. There are many example of this in our code.

src/components/query-panel/QueryPanelCtrl.ts
```diff
@@ -0,0 +1,74 @@
...
+          json: true
+        },
+        readOnly: true
+      };
+      this.editor = CodeMirror.fromTextArea(area, editorConfig);
+    }, 200);
+  };
+
+  getTemplateUrl(): string {
+    return "components/query-panel/view/" + this.panelType + "-panel.html";
```
ryanemerson: I think we can just use a generic template that handles all cases opposed to creating multiple panels.

src/module/cache/QueryCtrl.ts
```diff
@@ -0,0 +1,30 @@
...
+  currentCacheAvailability(): boolean {
+    return this.container.available;
+  }
+
+  currentClusterAvailabilityAsString(): string {
+    return this.container.available ? "AVAILABLE" : "N/A";
+  }
+
+  resetStats(): void {
+    this.cacheService.resetStats(this.container, this.cache).finally(() => {
```
ryanemerson: Can be one line `this.cacheService.resetStats(this.container, this.cache).finally(() => this.refresh());`

src/module/cache/view/query.html
```diff
@@ -0,0 +1,47 @@
...
+        </li>
+        <li>{{ctrl.container.profile}}</li>
+        <li><a ui-sref="container({profileName: ctrl.container.profile, containerName: ctrl.container.name})">{{ctrl.container.name}}</a>
+        <li class="active"><strong>Query</strong>
+      </ol>
+      <h1>
+                <span id-generator="availability" class="label"
+                      ng-class="ctrl.currentCacheAvailability() ? 'label-success' : 'label-danger'">
+                    {{ctrl.currentClusterAvailabilityAsString()}}
+                </span>&nbsp;{{ctrl.cache.name}}
```
ryanemerson: Indentation needs fixing.

src/services/rest/RestService.ts
```diff
@@ -0,0 +1,148 @@
...
+    };
+    let url: string = this.generateUrl(request);
+    return this.executeHelper("delete", url, request.jsonPayload, this.getDefaultHeaders());
+  }
+
+  executeCacheDeleteKey(cacheTarget: string, key: string): ng.IPromise<any> {
+    let request: IRestRequest = {
+      jsonPayload: undefined,
+      cache: cacheTarget,
+      urlSuffix: "/".concat(key)
```
ryanemerson: We should add the `/` in generateURL if the suffix is not undefined as currently you are repeating this many times.

src/services/rest/IRestRequest.ts
```diff
@@ -0,0 +1,5 @@
+export interface IRestRequest {
+  jsonPayload: any;
+  cache: string;
+  urlSuffix: string;
+}
```
ryanemerson: I think defining `IRestRequest` is overkill as it's only used internally by `RestService.ts`.

src/services/rest/RestService.ts
```diff
@@ -0,0 +1,148 @@
...
+      (success: any) => {
+        deferred.resolve(success);
+      },
+      (failure: any) => {
+        this.processResponseFailure(deferred, failure);
+      });
+    return deferred.promise;
+  }
+
+  private generateUrl(request: IRestRequest): string {
```
ryanemerson: I think this would be better if it simply took `jsonPayload`, `cache`, `urlSuffix` as three params. Cache is always required, whereas the other two appear to be optional. You can then just include the required params when calling `generateUrl` from the above helper methods.

src/services/rest/RestService.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  clearGetCache(): void {
+    this.$cacheFactory.get("$http").removeAll();
+  }
+
+  private executeHelper(type: string, url: string, data: any, headers?: ng.IRequestShortcutConfig): ng.IPromise<any> {
+    let deferred: ng.IDeferred<any> = this.$q.defer<any>();
+    let response: IHttpPromise;
+    switch (type) {
+      case "post":
```
ryanemerson: This would be much nicer if we could use enums introduced in Typescript 2.4. I know this PR isn't the place for such an upgrade, but we really should try to keep on top of updating our dependency versions as the longer we postpone such tasks, the harder these tasks become.

src/services/rest/RestService.ts
```diff
@@ -0,0 +1,148 @@
...
+      default:
+        deferred.reject("Unknown request type " + type);
+    }
+    response.then(
+      (success: any) => {
+        deferred.resolve(success);
+      },
+      (failure: any) => {
+        this.processResponseFailure(deferred, failure);
+      });
```
ryanemerson: Can be a one liner: `response.then(success => deferred.resolve(success), failure => this.processResponseFailure(deferred, failure));`

ryanemerson: In fact, `processResponseFailure` is only used here, so might as well do inline.

src/services/rest/RestService.ts
```diff
@@ -0,0 +1,148 @@
...
+    };
+    let url: string = this.generateUrl(request);
+    return this.executeHelper("post", url, request.jsonPayload, this.getDefaultHeaders());
+  }
+
+  clearGetCache(): void {
+    this.$cacheFactory.get("$http").removeAll();
+  }
+
+  private executeHelper(type: string, url: string, data: any, headers?: ng.IRequestShortcutConfig): ng.IPromise<any> {
```
ryanemerson: It seems that `getDefaultHeaders` is always passed to `executeHelper`, so we might as well just define the headers variable here instead of passing as a param.

src/components/query-panel/QueryPanelCtrl.ts
```diff
@@ -0,0 +1,74 @@
...
+      default:
+        console.log("Unknown request type " + type);
+    }
+    return response;
+  }
+
+  $postLink(): void {
+    this.$timeout(() => {
+      let area: HTMLTextAreaElement = <HTMLTextAreaElement>document.getElementById(this.name + "-response-text-area");
+      console.log("after timeout " + area + " for " + this.name);
```
ryanemerson: Remove this, it just creates noise in the console.

src/components/query-panel/view/delete-panel.html
vblagoje: +1 Ideally. I did not want to overoptimize right away. Something to strive to. Still waiting for UI feedback from Catherine. We can decide then what to do. But for the time being, I left the option to customize each screen individually.

src/components/query-panel/QueryPanelCtrl.ts
vblagoje: But of course it is. Making corrections. Thanks @ryanemerson

src/components/query-panel/QueryPanelCtrl.ts
vblagoje: Yes, definitely, but lets for Catherine.

src/module/cache/view/query.html
vblagoje: +1

src/services/rest/RestService.ts
vblagoje: +1

src/services/rest/IRestRequest.ts
vblagoje: +1 but it makes things cleaner. Lets remove it later when everything is done

src/services/rest/RestService.ts
vblagoje: Aha ok. You are right

vblagoje: I'll remove it now

src/services/rest/RestService.ts
vblagoje: +1

vblagoje:
@ryanemerson I addressed the most pressing issues so we can integrate it now. We'll polish everything even further after the initial integration

src/components/query-panel/QueryPanelCtrl.ts
```diff
@@ -0,0 +1,84 @@
...
+  statusCode: number;
+  editor: CodeMirror.Editor;
+  panelType:string;
+  inputParams: string [];
+
+  constructor(private restService: RestService, private $timeout: ng.ITimeoutService) {
+  }
+
+  executeAction(): void {
+    this.executionActionHelper(this.name, this.inputParams[0], this.inputParams[1]).then((response: any) => {
```
ryanemerson: As you are using class variables, there is no need to pass these values to `executionActionHelper`, instead you can just access them in the method.

src/components/query-panel/QueryPanelCtrl.ts
```diff
@@ -0,0 +1,84 @@
...
+export class QueryPanelCtrl implements ng.IComponentController {
+  static $inject: string[] = ["restService", "$timeout"];
+
+  cacheName: string;
+  name: string;
+  description: string;
+  statusCode: number;
+  editor: CodeMirror.Editor;
+  panelType:string;
+  inputParams: string [];
```
ryanemerson: Do you intend to have more than two inputs? If so an object would probably be a better choice than an array as we can then add meaningful keys to go with the stored value. If not then two appropriately named variables would suffice.

src/services/rest/RestService.ts
ryanemerson: This is still the case.

src/components/query-panel/QueryPanelCtrl.ts
vblagoje: +1, done

src/components/query-panel/QueryPanelCtrl.ts
vblagoje: Not sure at this point, maybe leave them for now until we see what the final number of params is?

src/services/rest/RestService.ts
ryanemerson: @vblagoje I think you edited my comment :smile: 

What I mean is that headers is always equal to `getDefaultHeaders`, so you might as well just define headers in the method once opposed to passing it everytime, i.e.:
```typescript
private executeHelper(type: string, url: string, data: any): ng.IPromise<any> {
let headers: ng.IRequestShortcutConfig = {
    headers: {
        "Accept": "*/*",
        "Content-type": "application/json"
    }
};
...
}
```

src/components/query-panel/QueryPanelCtrl.ts
ryanemerson: Sure

src/services/rest/RestService.ts
ryanemerson: @vblagoje This still needs doing.

vblagoje:
Corrected and verified @ryanemerson Give it a go

ryanemerson:
@vblagoje Integrated https://github.com/infinispan/infinispan-management-console/commit/fd4785faba15b1286ee6dba2ec5173d27a5aa1e3, thanks

ryanemerson:
@vblagoje FYI I have left the JIRA tickets open to reflect that this still needs some polish and is a WIP.

PR #236, opened 2018-04-02
ISPN-8460 Administration console - create user friendly configuration validation errors

vblagoje:
Master only. Please review @ryanemerson

### DIFF
--- a/src/services/dmr/DmrService.ts
+++ b/src/services/dmr/DmrService.ts
@@ -179,15 +179,43 @@ export class DmrService {
     let msg: string = "An unspecified error has been received from the server";
     if (response.status !== 401) {
       let result: any = response.data;
-      if (result && result["failure-description"] != null) {
-        if (isNotNullOrUndefined(result["failure-description"]["domain-failure-description"])) {
-          msg = result["failure-description"]["domain-failure-description"];
-        } else {
-          msg = result["failure-description"];
+      if (isNotNullOrUndefined(result)) {
+        if (isNotNullOrUndefined(result["failure-description"])) {
+          if (isNotNullOrUndefined(result["failure-description"]["domain-failure-description"])) {
+            msg = result["failure-description"]["domain-failure-description"];
+          } else {
+            msg = result["failure-description"];
+          }
+        }
+        // see if we can find the cause of this error
+        let cause: string = this.findErrorCause(result);
+        if (isNotNullOrUndefined(cause)) {
+          msg = cause;
         }
       }
       promise.reject(msg);
     }
+  }
+
+  private findErrorCause(root: any): string {
+    // traverse the object tree
+    let cause: string = undefined;
+    traverseObject(root, (key: string, value: any, trail: string []) => {
+      // NOTE "Operation step-1" contains the cause, adjust if necessary
+      let errorFound: string = this.visitTraversedErrorNode(value, "Operation step-1");
+      if (isNotNullOrUndefined(errorFound)) {
+        cause = errorFound;
+      }
+    });
+    return cause;
+  }
+
+  private visitTraversedErrorNode(obj: any, field: string): string {
+    let cause: string = undefined;
+    if (isNotNullOrUndefined(obj) && isNotNullOrUndefined(obj[field])) {
+      cause = obj[field];
+    }
+    return cause;
   }
 
   private addCompositeOperationsToBuilder(builder: CompositeOpBuilder, address: string[], config: any,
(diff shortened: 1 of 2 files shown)


src/services/dmr/DmrService.ts
```diff
@@ -179,17 +179,45 @@ export class DmrService {
...
+          if (isNotNullOrUndefined(result["failure-description"]["domain-failure-description"])) {
+            msg = result["failure-description"]["domain-failure-description"];
+          } else {
+            msg = result["failure-description"];
+          }
+        }
+        // see if we can find the cause of this error
+        let cause: string = this.findErrorCause(result);
+        if (isNotNullOrUndefined(cause)) {
+          msg = cause;
```
ryanemerson: This always overrides `msg` if it has previously been set. Why not look for the cause first. If not null, great you have your msg, otherwise check the failure description.

src/services/dmr/DmrService.ts
```diff
@@ -179,17 +179,45 @@ export class DmrService {
...
       promise.reject(msg);
     }
   }
 
+  private findErrorCause(root: any): string {
+    // traverse the object tree
+    let cause: string = undefined;
+    traverseObject(root, (key: string, value: any, trail: string []) => {
+      // NOTE "Operation step-1" contains the cause, adjust if necessary
+      let errorFound: string = this.visitTraversedErrorNode(value, "Operation step-1");
```
ryanemerson: No need for the variable `errorFound` or method `visitTraversedErrorNode` as only used once and just adds verbosity. Can be:

```typescript
if (isNotNullOrUndefined(value) && isNotNullOrUndefined(obj["Operation step-1"]) {
    cause = obj["Operation step-1"];
}
```

src/services/dmr/DmrService.ts
```diff
@@ -179,17 +179,45 @@ export class DmrService {
...
         }
       }
       promise.reject(msg);
     }
   }
 
+  private findErrorCause(root: any): string {
+    // traverse the object tree
+    let cause: string = undefined;
+    traverseObject(root, (key: string, value: any, trail: string []) => {
```
ryanemerson: Can be simplified to `traverseObject(root, (key, value, trail) => {`

Also, as you are using `traverseObject` does that not mean that it's possible for the cause to be set multiple times with the current code? Don't we just want to return after finding the first cause?

src/services/dmr/DmrService.ts
```diff
@@ -179,17 +179,45 @@ export class DmrService {
...
+      if (isNotNullOrUndefined(result)) {
+        if (isNotNullOrUndefined(result["failure-description"])) {
+          if (isNotNullOrUndefined(result["failure-description"]["domain-failure-description"])) {
+            msg = result["failure-description"]["domain-failure-description"];
+          } else {
+            msg = result["failure-description"];
+          }
+        }
+        // see if we can find the cause of this error
+        let cause: string = this.findErrorCause(result);
```
ryanemerson: Similar to `visitTraversedErrorNode` I'm not sure what benefit we get from having this as a standalone method.

src/services/dmr/DmrService.ts
ryanemerson: Also, what happens if the error is on `Operation step-<x>`

src/services/dmr/DmrService.ts
vblagoje: +1 ok, makes sense

src/services/dmr/DmrService.ts
vblagoje: Yes, we do. It does not matter much tbh. We don't have abort the loop mechanism in traverseObject...how would you abort looping?

src/services/dmr/DmrService.ts
vblagoje: +1

src/services/dmr/DmrService.ts
vblagoje: True, how would we do the search for other Operation step or if the op was not composite. I'll investigate this a bit further

src/services/dmr/DmrService.ts
ryanemerson: You could modify `traverseObject` so that it accepts a callback which returns true once the desired state has been reached. If the callback returns false, keep traversing, otherwise stop.

vblagoje:
I have further simplified the error search algorithm. It looks for fields by regex pattern in the traversed error objects received. When it finds such a field it grabs the value and returns it as an error. If nothing is found then the default error is thrown. Tried it with several errors and it works nicely. Have a look @ryanemerson

ryanemerson:
@vblagoje Damm, I forgot to squash the "PR review" commit before rebasing and I can't force push to master.

PR #239, opened 2018-04-06
Master response fix

vblagoje:
@ryanemerson hopefully the last one. Please review

### DIFF
--- a/src/components/query-panel/QueryPanelCtrl.ts
+++ b/src/components/query-panel/QueryPanelCtrl.ts
@@ -19,20 +19,28 @@ export class QueryPanelCtrl implements ng.IComponentController {
 
   executeAction(): void {
     this.executionActionHelper().then((response: any) => {
-      this.statusCode = 200;
+      if (isNotNullOrUndefined(response.status)) {
+        this.statusCode = response.status;
+      } else {
+        this.statusCode = 200;
+      }
       if (isNotNullOrUndefined(response) && isNotNullOrUndefined(response.data)) {
         if (isObject(response.data)) {
           this.editor.setValue(JSON.stringify(response.data, null, "\t"));
         } else {
-          this.editor.setValue("Request executed successfully. No response returned.");
+          this.editor.setValue(response.statusText);
         }
       }
     }).catch((error: any) => {
-      this.statusCode = 401;
-      if (isEmptyString(error)) {
-        this.editor.setValue("Error response with no further details returned");
+      if (isNotNullOrUndefined(error.status)) {
+        this.statusCode = error.status;
       } else {
-        this.editor.setValue(JSON.stringify(error, null, "\t"));
+        this.statusCode = 401;
+      }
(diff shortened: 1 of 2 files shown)


src/components/query-panel/QueryPanelCtrl.ts
```diff
@@ -19,20 +19,28 @@ export class QueryPanelCtrl implements ng.IComponentController {
 
   executeAction(): void {
     this.executionActionHelper().then((response: any) => {
-      this.statusCode = 200;
+      if (isNotNullOrUndefined(response.status)) {
```
ryanemerson: The response should never be null here. If `executionActionHelper` goes into the default switch statement, we should reject the promise as well as writing an error. That way this part of the code will never be executed with a null response object. Furthermore, is it every possible for a response object not to have a status? I assume that in the event of a server error a status of `5xx` would be returned at least.

src/components/query-panel/QueryPanelCtrl.ts
```diff
@@ -19,20 +19,28 @@ export class QueryPanelCtrl implements ng.IComponentController {
...
+        this.statusCode = response.status;
+      } else {
+        this.statusCode = 200;
+      }
       if (isNotNullOrUndefined(response) && isNotNullOrUndefined(response.data)) {
```
ryanemerson: If `response.data` is null, it might still be possible to utilise `statusText`.

src/components/query-panel/QueryPanelCtrl.ts
```diff
@@ -19,20 +19,28 @@ export class QueryPanelCtrl implements ng.IComponentController {
...
-          this.editor.setValue("Request executed successfully. No response returned.");
+          this.editor.setValue(response.statusText);
         }
       }
     }).catch((error: any) => {
-      this.statusCode = 401;
+      if (isNotNullOrUndefined(error.status)) {
+        this.statusCode = error.status;
+      } else {
+        this.statusCode = 401;
```
ryanemerson: Why would this be 401? The problem is not related to authorization. See https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

src/services/rest/RestService.ts
```diff
@@ -87,13 +87,7 @@ export class RestService {
...
         deferred.reject("Unknown request type " + type);
     }
-    response.then(
-      (success: any) => { deferred.resolve(success); },
-      (failure: any) => {
-        if (failure.status !== 401) {
-          deferred.reject(failure.data);
-        }
-      });
+    response.then((success: any) => { deferred.resolve(success); }, (failure: any) => { deferred.reject(failure); });
```
ryanemerson: can be:
```
response.then(success => deferred.resolve(success), failure => deferred.reject(failure));
```

ryanemerson:
@vblagoje Integrated https://github.com/infinispan/infinispan-management-console/commit/da8d80f2584037c47098687010379be4df127569, thanks

PR #243, opened 2018-04-11
ISPN-9066 Successfully configured Slave node with no server configured is not shown in the list of hosts

vblagoje:
Please review @ryanemerson

### DIFF
--- a/src/module/server-group/ServerGroupCtrl.ts
+++ b/src/module/server-group/ServerGroupCtrl.ts
@@ -21,13 +21,12 @@ import {ModalService} from "../../services/modal/ModalService";
 
 export class ServerGroupCtrl {
   static $inject: string[] = ["$state", "$uibModal", "dmrService", "serverGroupService", "serverService",
-    "jGroupsService", "launchType", "serverGroup", "available", "status", "$filter", "modalService"];
+    "jGroupsService", "launchType", "serverGroup", "available", "status", "hosts", "$filter", "modalService"];
 
   serverStatusMap: IMap<string> = {};
   statuses: string [] = [];
   serverInetMap: IMap<string> = {};
   coordinator: IServerAddress;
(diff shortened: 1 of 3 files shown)


src/module/server-group/ServerGroupCtrl.ts
```diff
@@ -228,10 +230,10 @@ export class ServerGroupCtrl {
     return this.statuses;
   }
 
-  private filterUniqueHosts(): string[] {
-    return this.serverGroup.members
-      .map((server) => server.host)
-      .filter((item, post, array) => array.indexOf(item) === post);
+  private fetchHosts(): string[] {
+    this.domainService.getHosts().then((hosts) => {
```
ryanemerson: We should probably resolve this in `ServerGroup.ts` and pass to the controller.

src/services/domain/DomainService.ts
```diff
@@ -38,6 +38,22 @@ export class DomainService {
...
+      address: [],
+      "child-type": "host"
+    };
+    let hosts: string [] = [];
+    this.dmrService.readChildResources(request).then((response) => {
+      for (let hostname in response) {
+        hosts.push(hostname);
+      }
+      deferred.resolve(hosts);
+    });
```
ryanemerson: I think we can make use of `Object.keys` here and turn this into a one-liner, something like:

```typescript
this.dmrService.readChildResources(request).then(response => deferred.resolve(Object.keys(response)));
```

vblagoje:
Updated, thanks @ryanemerson

vblagoje:
Hey @ryanemerson is this one gtg, anything else we need to do?

ryanemerson:
@vblagoje LGTM, can you just squash the commits and re-push please.

vblagoje:
@ryanemerson np, squashed and forced pushed

PR #244, opened 2018-04-20
ISPN-9039 Parametrized admin console branding

vblagoje:
@ryanemerson just a preview of how we can do all the string related branding. Comments and feedback welcome. We'll deal with images separately


gulpfile.js
```diff
@@ -19,7 +19,9 @@ const config = {
     typingsDir: path.join(__dirname, 'typings'),
     targetDir: path.join(__dirname, 'target'),
     fonts: path.join('fonts'),
-    moduleName: 'app'
+    moduleName: 'app',
+    assetsRepo: 'https://github.com/vblagoje/infinispan-management-console-assets.git',
```
vblagoje: @ryanemerson this is just a test repo. We'll use a proper one when PR review is done

vblagoje:
@ryanemerson @tristantarrant this is ready to go. We need to define repo for assets - I am currently using a test repo. The workflow is fully automated and does not require human intervention. @ryanemerson please have a look when you get a chance

tasks/add-assets-repo.js
```diff
@@ -0,0 +1,9 @@
+module.exports = (gulp, config) => () => {
```
ryanemerson: AFAICT, this script is never used?

tasks/merge-assets.js
```diff
@@ -0,0 +1,8 @@
+module.exports = (gulp, config) => () => {
+  const git = require('gulp-git');
+
+  return git.pull('assets', config.branding, {args: '--squash --allow-unrelated-histories'}, function (err) {
+    if (err) throw err;
+  })
```
ryanemerson: missing `;`

tasks/constants.js
```diff
@@ -0,0 +1,22 @@
...
+  const path = require('path');
+  const rename = require('gulp-rename');
+  var ngConstant = require('gulp-ng-constant');
+  var config = require('../src/constants.json'), envConfig = config.infinispan;
+
+
+  if (cfg.branding === 'infinispan') {
+    envConfig = config.infinispan;
+  }
+  else {
```
ryanemerson: Should be `} else {`

ryanemerson:
With the translate approach you can also use your build strips with regards to assets to simply pull the "local-enUS.txt" file from the other repo to replace the infinispan default.

vblagoje:
Ok @ryanemerson I see your point. I don't like that we lose "strong typing" we have with Constants.ts approach. We might have entries that miss providing translation for. With translate approach, we can only check omissions at runtime - visually.

ryanemerson:
@vblagoje good point and also I didn't take into account that `brandImage` is needed for `ng-source={{}}`, so I'm not sure translate would actually work in that scenario.

tasks/add-assets-repo.js
vblagoje: +1 it is used to execute "git remote add" - it will be used by automated builds as we need to add assets repo to fetch a particular assets branch from it. Devs would likely invoke this only once

vblagoje:
@ryanemerson yeah, it might not work as both values are resolved at runtime. So we go with Constant.ts solution approach?

vblagoje:
Ok @ryanemerson let's just consult with @tristantarrant about the repo we'll use for resources. I don't have privileges to create repos.

ryanemerson:
@vblagoje If you squash these commits it's good to go once you have an answer to the repo question.

vblagoje:
@ryanemerson I've added --repo and --branch parameters. So one would invoke "gulp add-assets-repo --repo https_path_to_repo" and then invoke "gulp merge-assets --branch infinispan" where infinispan is the assets branch on the given repo. Lmk your thoughts

ryanemerson:
@vblagoje With these changes what would be the one-liner to build a deployable jar using a custom assets repo?

vblagoje:
@ryanemerson hmmm, good point. let me try and get back to you.

vblagoje:
@ryanemerson I made this work. Try it out on your local branch directly:

gulp brand --repo https://github.com/vblagoje/infinispan-management-console-assets.git --branch infinispan

tasks/brand.js
```diff
@@ -0,0 +1,12 @@
+module.exports = (gulp, config) => () => {
+
+  const git = require('git-promise');
+  return git('remote add assets ' + config.cmdParams.repo).then(function () {
+    return git('pull assets ' + config.cmdParams.branch + ' --squash --allow-unrelated-histories').then(function(){
+      return git('add ./assets').then(function(){
+        return git('commit --m \'Updated assets with branding \'' + config.cmdParams.branch);
```
ryanemerson: Will this not add the commit to the "infinispan-management-console" repo? Or is on the remote repo?

pom.xml
```diff
@@ -112,14 +112,34 @@
...
+                        <goal>gulp</goal>
+                      </goals>
+
+                      <!-- optional: the default phase is "generate-resources" -->
+                      <phase>generate-resources</phase>
+
+                      <configuration>
+                        <!-- optional: if not specified, it will run gulp's default
+                        task (and you can remove this whole <configuration> section.) -->
+                        <arguments>brand --repo ${branding.repo} --branch ${branding.branch}</arguments>
```
ryanemerson: Can we not just have this one execution as the default? We could define branding.rep and branding.branch in the maven repo and then if we want to pull from a different repo/branch we just specify `mvn clean install -Dbranding.repo=blah`

vblagoje:
@ryanemerson this should be close to it - what we want. Give it a look

bin/utils.py
```diff
@@ -375,15 +375,20 @@ def upload_rsync(self, fr, to, flags = []):
...
     
   for c in mvn_commands:
+    mvn_command = c[0]
     if settings['dry_run']:
       c.append("-DskipNexusStagingDeployMojo=true -Dmaven.deploy.skip=true")
     if not settings['verbose']:
       c.insert(0, '-q')
+    if mvn_command == 'install' and args.assetRepo and args.assetBranch: 
+      mvn_param = '-Dbranding.repo={assetRepo} -Dbranding.branch={assetBranch}'
+      mvn_param = mvn_param.format(assetRepo=args.assetRepo, assetBranch=args.assetBranch)
```
ryanemerson: This can be a oneliner and you can just use `{}` as only two args, so: `"-Dbranding.repo={} -Dbranding.branch={}".format(args.assetRepo, args.assetBranch)`. Also note that we should use double quotes.

bin/utils.py
```diff
@@ -375,15 +375,20 @@ def upload_rsync(self, fr, to, flags = []):
   def upload(self, fr, to, type):
     self.copy(fr, "%s/%s/%s" % (self.location_root, type, to))    
 
-def maven_build_distribution(version):
+def maven_build_distribution(version, args):
```
ryanemerson: I think we should pass `assetRepo` and `assetBrand` directly here as this is a generic util method and should not be tightly coupled to argsparse.

pom.xml
vblagoje: Ok, it's a deal

bin/utils.py
vblagoje: Ok +1

bin/utils.py
vblagoje: +1

pom.xml
vblagoje: Yeah, let's but at the moment we don't have those branches. It is easy to add them as properties in pom.xml for both products. We yet have to agree on how we are going to do that. In the meantime, I made changes so that branding only occurs if repo and branch are specified. Until then then build stays essentially the same. I'll commit that change now

package.json
```diff
@@ -42,6 +42,7 @@
     "karma-mocha-reporter": "^2.0.1",
     "karma-phantomjs-launcher": "~1.0.0",
     "less": "2.7.1",
+    "minimist": "^1.2.0",
```
vblagoje: @ryanemerson I forgot this dependency

pom.xml
ryanemerson: So at the moment when we call `mvn clean install`, both `gulp brand` and then `gulp build` are being called. We should just need to call `gulp build`. In `gulpfile.js` we can simply add `brand` as one of the requirements for the `build` task. If no `cmdParams.repo` or `cmdParams.branch` is specified, then do nothing (the current code in `brand.js`). In the pom.xml, we can set the default value of `branding.repo` and `branding.branch` to be empty, so unless these parameters are set no branding with external repos occur when `mvn clean install` is called.

vblagoje: Ok +1

ryanemerson:
@vblagoje LGTM, can you rebase please

vblagoje:
Rebased @ryanemerson

PR #245, opened 2018-05-03
ISPN-8989 Administration console - Counters don't work in standalone mode

vblagoje:
@ryanemerson we can merge this one into 9.2 and master, of course.

### DIFF
--- a/src/module/cache-container/view/cache-container.html
+++ b/src/module/cache-container/view/cache-container.html
@@ -75,7 +75,7 @@
     <div class="col-sm-12 col-md-12">
       <ul class="nav nav-tabs">
         <li ui-sref-active="active" ui-sref=".caches"><a translate>CACHES</a></li>
-        <li ui-sref-active="active" ui-sref=".counters" ng-if="!ctrl.isLocalMode()"><a translate>COUNTERS</a></li>
+        <li ui-sref-active="active" ui-sref=".counters"><a translate>COUNTERS</a></li>
         <li ui-sref-active="active" ui-sref=".tasks" ng-if="!ctrl.isLocalMode()"><a translate>TASK_EXECUTION</a></li>
       </ul>
     </div>
--- a/src/services/counters/CounterService.ts
+++ b/src/services/counters/CounterService.ts
@@ -52,40 +52,52 @@ export class CounterService {
   }
 
   getAllCountersHelper(cacheContainer: ICacheContainer): ng.IPromise<ICounter[]> {
-
+    let request: IDmrRequest;
     let deferred: ng.IDeferred<ICounter[]> = this.$q.defer<ICounter[]>();
-    this.jGroupsService.getServerGroupCoordinator(cacheContainer.serverGroup).then((coord: IServerAddress) => {
-
-      let request: IDmrRequest = <IDmrRequest>{
-        address: this.getCountersRuntimeAddress(cacheContainer.name, coord.host, coord.name),
+    if (this.launchType.isStandaloneMode()) {
+      request = <IDmrRequest>{
+        address: this.getCountersRuntimeAddress(cacheContainer.name),
         "recursive-depth": 2,
         "include-runtime": true
       };
-
-      this.dmrService.readResource(request)
-        .then((response: any): ICounter[] => {
-          let counters: ICounter[] = [];
-          let trail: String [] = [];
-          traverse(response, (key: string, value: string, trail: string []) => {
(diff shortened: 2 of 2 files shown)


src/module/cache-container/view/cache-container.html
```diff
@@ -75,7 +75,7 @@ <h1><span class="label"
     <div class="col-sm-12 col-md-12">
       <ul class="nav nav-tabs">
         <li ui-sref-active="active" ui-sref=".caches"><a translate>CACHES</a></li>
-        <li ui-sref-active="active" ui-sref=".counters" ng-if="!ctrl.isLocalMode()"><a translate>COUNTERS</a></li>
+        <li ui-sref-active="active" ui-sref=".counters"><a translate>COUNTERS</a></li>
```
ryanemerson: This is strange, even though this looks correct when I load up the server with `standalone.sh` no counters tab is present.

src/services/counters/CounterService.ts
```diff
@@ -102,12 +114,19 @@ export class CounterService {
...
   reset(container: ICacheContainer, counter: ICounter): ng.IPromise<any> {
-    return this.jGroupsService.getServerGroupCoordinator(container.serverGroup).then((coord: IServerAddress) => {
+    if (this.launchType.isStandaloneLocalMode()) {
       return this.dmrService.executePost({
-        address: this.getCounterRuntimeAddress(counter, container.name, coord.host, coord.name),
+        address: this.getCounterRuntimeAddress(counter, container.name),
         operation: "counter-reset"
       });
-    });
+    } else {
```
ryanemerson: You don't handle `isStandaloneClusteredMode` here which I think is why it fails with standalone.sh -c clustered.xml

src/services/counters/CounterService.ts
```diff
@@ -52,42 +52,54 @@ export class CounterService {
...
-            }
-          }, trail);
-          return counters;
-        }).then(counters => {
+      this.findCounters(request).then(counters => {
         deferred.resolve(counters);
       });
-    });
+    } else {
+      this.jGroupsService.getServerGroupCoordinator(cacheContainer.serverGroup).then((coord: IServerAddress) => {
```
ryanemerson: No need for typing here, can just be `coord => {`

src/services/counters/CounterService.ts
```diff
@@ -52,42 +52,54 @@ export class CounterService {
...
+        this.findCounters(request).then(counters => {
+          deferred.resolve(counters);
+        });
+      });
+    }
     return deferred.promise;
   }
 
+  findCounters(request: IDmrRequest): ng.IPromise<ICounter[]> {
+    return this.dmrService.readResource(request).then((response: any): ICounter[] => {
```
ryanemerson: No need for typing on lambda

src/module/cache-container/view/cache-container.html
vblagoje: Try using incognito windows...I'll try again myself.

src/services/counters/CounterService.ts
vblagoje: I'll check - it works for me. Don't forget that you also need to remove counters=COUNTERS on fresh restart of the test between standalone and clustered

src/services/counters/CounterService.ts
vblagoje: Ok, I see what you mean. It is easier for me to think about the object, functions, and code in general if it's type is declared.

src/services/counters/CounterService.ts
vblagoje: ok

vblagoje:
Ok @ryanemerson I'll recheck everything once again. Some errors might have slipped by.

src/services/counters/CounterService.ts
ryanemerson: Why? If you are starting the server with different configurations then the state is persisted to the currently loaded config. So starting the server with `standalone.sh` and `standalone.sh -c clustered.xml` should be separate.

ryanemerson:
@vblagoje Note I am running the serve with `gulp serve:dist` which is how the console behaves once distributed, so that might explain why we're experiencing different things. 
- `standalone.sh` counters tab is still not present
- `standalone.sh -c clustered.xml` now works :+1:

ryanemerson:
@vblagoje Ignore my last comment, running `gulp clean` explicitly solved the problem. Thanks

PR #247, opened 2018-05-16
ISPN-8764 Add schema editor

vblagoje:
Master only. Update schema editor with line numbers and syntax highlighting.


vblagoje:
@ryanemerson checking. Good catch.

vblagoje:
Rebased. schemaForm.schemaBody.$valid is not valid after we dress it up with CodeMirror. Not sure how we can handle this scenario tbh. The consequences are not bad. We let people submit schemas that are empty (i.e. empty editor).

vblagoje:
@ryanemerson ^^

vblagoje:
Yeah @ryanemerson, I tried already this.editor.getValue() but I got
nowhere. You are right, I'll try a bit harder tomorrow.

vblagoje:
@ryanemerson not sure what to do here. I capture clicks from the editor and check if it has content. That's easy. However, it seems that the angular cycle that checks ng-disabled is not getting invoked because we wrapped text area with CodeMirror editor. Any ideas?

vblagoje:
@ryanemerson see "non working solution" commit

ryanemerson:
Why do we need to capture clicks? It's typing that is important. Can we not just check the content of `this.editor.getValue();` in a regular controller method. The template will automatically poll this as angular normal does.

vblagoje:
The goal is to enable button "Create Schema" if there is content in the editor, right? Unless you put your cursor outside of the editor into the text field that contains the name of the schema file Angular does not check ng-disabled. Angular stops doing ng-disabled checks after we dress up text area with CodeMirror.Editor...

PR #250, opened 2018-06-04
ISPN-7291 It should be possible to delete cache from Administration console

vblagoje:
Master only, do not integrate yet. There are still some issues with standalone setup. Might be transient, investigating.


ryanemerson:
Are there still issues with this @vblagoje ?

vblagoje:
@ryanemerson I'll try again with the latest server. These are backend problems, not the front end.

ryanemerson:
Any update?

vblagoje:
Let me try it now @ryanemerson

vblagoje:
@ryanemerson let's integrate it in master - it works. And if we don't encounter any issues we can soon port it back as well

PR #251, opened 2018-06-18
Generate license information

jsenko:
Using this set of scripts, XSL templates, and data files, a Maven artifact containing license information for all production dependencies is generated.


ryanemerson:
Thanks @jsenko

PR #253, opened 2018-06-28
ISPN-9331 Cluster node page flickers angular ctrl variables

vblagoje:
Master, 9.1.x and 9.2.x.


ryanemerson:
@vblagoje We don't support 9.2.x or 9.1.x now that 9.3.x has been  released. What we can do to prevent a backport though, is wait until we merge this PR into master before branching for 9.3.x

PR #252, opened 2018-06-28
ISPN-9256 Upgrade npm dependencies

vblagoje:
Carefully review at your leisure moment @ryanemerson I tried 'mvn install' on my fedora laptop from scratch and it worked. No rush to integrate but we should upgrade at the appropriate moment in dev cycle.

PR #256, opened 2018-06-29
ISPN-9337 Lock admin console dependencies

vblagoje:
For 9.3.x branch


vblagoje:
This one is ok, verified with full rebuild

ryanemerson:
This results in `src/jspm.conf.js` having unstaged changes after the build

vblagoje:
@ryanemerson this does not happen to me. I clean all directories and even after fresh install - there are no changes to src/jspm.conf.js I thought it could be node version related. I built it with two different versions of the node without any issues. Could it be OS platform specific?

ryanemerson:
@vblagoje That comment was 10 days ago ... can you try what I suggested yesterday on https://github.com/infinispan/infinispan-management-console/pull/252 please. I suspect the cause of the dirty git repo is the same.

Also, are we keeping this PR or close in favour of https://github.com/infinispan/infinispan-management-console/pull/258 which also locks the dependencies?

vblagoje:
Yeah, I would just integrate #258 , we can close this one. I'll try it on #252 now

PR #257, opened 2018-07-11
Update proxy settings for WildFly 13

vblagoje:
Master only. Please verify @ryanemerson


ryanemerson:
Thanks @vblagoje. We still have the issue that Angular appears to be loaded twice, resulting in duplicate server calls, but seen as we now have a working dev environment I have created https://issues.jboss.org/browse/ISPN-9360 to track the issue.

vblagoje:
Ok Ryan - I'll take care of it now

PR #258, opened 2018-07-11
T 9360 for 9.3.x

vblagoje:
ISPN-9360 Angular Loaded Twice for 9.3.x along with sanitization fix as well


vblagoje:
@ryanemerson there is an extra commit for locking dependencies here. I forgot we did not integrate the previous PR

PR #259, opened 2018-08-28
ISPN-9642 Management console displays nothing after enabling server hinting

vblagoje:
See more details at https://issues.jboss.org/browse/ISPN-9462


vblagoje:
@arnav300 tried this branch on the product side - for the actual case where he encountered the problem

PR #263, opened 2018-10-26
Improving UX when user wants delete memory configuration

gustavolira:
When user wants delete memory (or any other) configuration is a little bit confusing.
First of all, user may thing his configuration could be deleted immediately (no confirmation when you click at a button). Currently to delete user need click at delete icon and after that click on Apply changes (far from delete icon) 

Second user can think that configuration was already deleted just clicking at delete icon and go to another page (link) and when they return to memory configuration they realize that configuration wasn't deleted.

With this commit user is asked if they really want delete the configuration (image attached) and when they confirm the configuration will automatically update and user don't need to click on "Apply changes" button
screen shot 2018-10-26 at 16 23 57

### DIFF
--- a/src/components/memory/MemoryCtrl.ts
+++ b/src/components/memory/MemoryCtrl.ts
@@ -6,6 +6,8 @@ import {
   makeFieldClean,
   makeFieldDirty
 } from "../../common/configuration/ConfigUtil";
+import {IModalService} from "angular-ui-bootstrap";
+import {openConfirmationModal} from "../../common/dialogs/Modals";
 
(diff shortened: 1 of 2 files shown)


gustavolira:
if this improvement is a good idea, I can replicate it to all other configuration items

ryanemerson:
@gustavolira I think making it so that the delete button applies the update after a confirmation is a good idea, however we will also then need to display the restart server modal (if required for the delete configuration) so that the change can take effect.

ryanemerson:
Closing as master is no longer actively developed

PR #275, opened 2019-02-07
ISPN-9952 Admin console hangs after cache create

vblagoje:
For master and other affected branches. I could have solved this one perhaps a bit differently (simpler) but I wanted to make something that could be reused in the future when we face these kinds of situations.

### DIFF
--- a/src/services/cache-config/CacheConfigService.ts
+++ b/src/services/cache-config/CacheConfigService.ts
@@ -375,7 +375,7 @@ export class CacheConfigService {
   private createCacheLoader(builder: CompositeOpBuilder, address: string[], config: any): void {
     let loaderClass: string = deepGet(config, "loader.LOADER.class");
     if (isNotNullOrUndefined(loaderClass) && loaderClass !== "None") {
-      this.createHelper(builder, address.concat("loader", "LOADER"), config.loader);
+      this.createHelper(builder, address.concat("loader", "LOADER"), config.loader, true);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/services/cache-config/CacheConfigService.ts
```diff
@@ -375,7 +377,7 @@ export class CacheConfigService {
   private createCacheLoader(builder: CompositeOpBuilder, address: string[], config: any): void {
     let loaderClass: string = deepGet(config, "loader.LOADER.class");
     if (isNotNullOrUndefined(loaderClass) && loaderClass !== "None") {
-      this.createHelper(builder, address.concat("loader", "LOADER"), config.loader);
+      this.createHelper(builder, address.concat("loader", "LOADER"), config.loader, true);
```
ryanemerson: `createHelper` only has three params, not sure why you're passing a boolean here

src/services/cache-config/CacheConfigService.ts
```diff
@@ -387,7 +389,7 @@ export class CacheConfigService {
 
     let objectKey: string = storeType.toUpperCase().replace(/-/g, "_");
     config[storeType][objectKey]["required-node"] = true;
-    this.createHelper(builder, address.concat(storeType, objectKey), config[storeType]);
+    this.createHelper(builder, address.concat(storeType, objectKey), config[storeType], true);
```
ryanemerson: `createHelper` only has three params, not sure why you're passing a boolean here

src/services/cache-config/CacheConfigService.ts
```diff
@@ -345,12 +346,16 @@ export class CacheConfigService {
     return deferred.promise;
   }
 
-  private createHelper(builder: CompositeOpBuilder, address: string[], config: any): void {
+  private createHelper(builder: CompositeOpBuilder, address: string[], config: any, requiredNode?: boolean): void {
```
vblagoje: @ryanemerson I changed the createHelper function slightly to include optional parameter requiredNode. If true, we'll add required-node field to a config object. See below

ryanemerson:
Closing as I have rebased these commits and slightly changed the approach: https://github.com/infinispan/infinispan-management-console/commit/103b0dd95cbd266cea569f32d20f8ccbf2eb02a6. Thanks @vblagoje

vblagoje:
Sure @ryanemerson Thanks for pushing these last few commits through the finish line!

PR #276, opened 2019-02-08
[ISPN-9955] Admin console - clusters link does not work in standalone mode

vblagoje:
Both master and 9.x branch. See https://issues.jboss.org/browse/ISPN-9955 for more details


ryanemerson:
Thanks @vblagoje, I adapted your commit so that it uses the shorthand if and inline lamda: https://github.com/infinispan/infinispan-management-console/commit/e5b99582ad7a77aac5c523f7cb55f30faab162a3

